PR #21, opened 2019-07-30
Updated the pom.xml to allow CFR to compile to a single JAR

ThisIsLibra:
This change makes it possible to compile CFR into a single JAR by using "mvn clean compile assembly:single". Additionally, the XML file has been beautified.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,13 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-
     <groupId>org.benf</groupId>
     <artifactId>cfr</artifactId>
     <version>0.147-SNAPSHOT</version>
-
     <name>cfr</name>
     <description>CFR Java decompiler</description>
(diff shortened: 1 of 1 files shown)


leibnitz27:
CFR doesn't have any dependencies - what does this buy you over 'mvn clean package'?
(my maven-fu is limited, honest question).

ThisIsLibra:
My understanding of Maven isn't too great either, so I might be mistaken here, but this is what I thought:

The change allows for an additional build method, where the project is build into a JAR (including all dependencies) without the sources or Javadoc. 

Additionally, I am unable to compile the project with the command you gave, as the Javadoc does not accept the "-html5" argument that is passed to it (openjdk8 on my machine), which might be related to my machine, but I am able to compile it using the command I added in this PR.

Does that provide an answer to your question?

leibnitz27:
Ah - ok, so this is html5 support being added for javadoc in j9.   I'd honestly prefer not to go down the 'add all dependencies' route you outline in this PR, because it implies I'm happy with dependencies, and I don't want to have the 'this dependency doesn't work in j6' conversation again and again ;)

Can you instead try dropping the '-html5' argument from the javadoc stanza in the pom?   This should then allow 'mvn package' to behave, and generate

31/07/2019  10:15           435,607 cfr-0.147-SNAPSHOT-javadoc.jar
31/07/2019  10:15           855,885 cfr-0.147-SNAPSHOT-sources.jar
31/07/2019  10:15         1,749,250 cfr-0.147-SNAPSHOT.jar

ThisIsLibra:
That works exactly as you described it would. Would it be an idea to exclude the '-html5' tag from the POM by default? I'd love to embed this repository into AndroidProjectCreator, but Java 8 is the maximum version there, meaning that the project cannot be build without breaking other dependencies there. As the build process is automatic, changing the POM locally isn't the best idea.

I fully understand it if you do not want to make that change, and I have no problem with that. In any case, I'd like to thank you for your work on this amazing tool!

leibnitz27:
That's fine by me.  It's a very small change, so I've just done it, save you making another PR.

ThisIsLibra:
Thank you very much!

leibnitz27:
FYI (not a problem) https://github.com/leibnitz27/cfr/issues/22

leibnitz27:
Tweaked a bit further to selectively enable flag in j9+.

PR #67, opened 2019-12-11
Improved Literal Formatting

skyrising:
Improves formatting of float and double literals obtained from casting or using `Math.PI`.

This will improve readability of code using float constants in double calculations (since for example `(double) 0.1f == 0.10000000149011612`) and trigonometry code (since for example `Math.PI / 180 == 0.017453292519943295`)

Applies the hex formatting code from longs to ints and raises the limit for number of different hex digits to 3 to facilitate the common use case of hex for rgb colors.

Test case:
```java
package org.benf.cfr.tests;

public class TypedLiteral1 {
    public static final double D_INF = Double.POSITIVE_INFINITY;
    public static final float F_INF = Float.POSITIVE_INFINITY;
    public static final double POINT_1 = 0.1f;
    public static final double POINT_01 = 0.01f;
    public static final double D_PI = Math.PI;
    public static final double D_PI_180 = Math.PI / 180;
    public static final double D_PI_3 = 3 * Math.PI;
    public static final float F_PI = (float) Math.PI;
    public static final float F_PI_180 = (float) (Math.PI / 180);
    public static final float F_PI_3 = (float) (3 * Math.PI);
    public static final int COLOR_WHITE = 0xffffff;
    public static final int COLOR_YELLOW_1 = 0xFFAA00;
    public static final int COLOR_YELLOW_2 = 0xE5E533;
}
```

Existing tests are unaffected (Tested 8, 10, 12 and ECJ 8)


leibnitz27:
Hey.  There's some lovely ideas in here.

I'm not quite sure about having a bunch of operations as pure strings - it feels wrong, though I get what you mean - it would certainly affect any consumer that takes the output as a token stream (I know there's at least one).  [ I know, I was doing this for the Inf case.  Hypocrite! But this makes the motivation to do it properly a lot stronger. ]

I think at this point, it would probably be worth splitting out literal transformation like this into an expressionRewriter pass - that way we get a proper AST internally.

@skyrising - do you fancy having a go at that? If not, I'll pick up the idea in this CR, but may hack it around a fair bit before merging.

I'm also *mildly* concerned about         

```
        float nearestFloat = (float) d;
        if (Double.compare(nearestFloat, d) == 0) {

```

which I _absolutely can't_ demonstrate is dodgy, but is setting off alarms ;)
(and I can't trigger it - would naively expect  `static double x2 = ((float)(Math.PI / 90));` to do so, but no joy.).

(Is this common practice?  We have the type, and I can't imagine that many people upcast a constant expression of pi).

skyrising:
Yeah, I also had the feeling that having a proper AST would be nice, I'll give it a shot and see how it goes.

I'm pretty confident that `Double.compare(f, (double) f) == 0` for all floats since, to my knowledge, all floats can be represented as double without loss of precision.

As far as common practice goes, I've seen lots of instances of this happening in the Minecraft code base. Most likely it's caused by implicit, rather than explicit casting.

leibnitz27:
Sure thing - if any of it is overly idiosyncratic / confusing then it's totally on me ;). 

Best to derive AbstractExpressionRewriter, then just test for literal and rewrite with your existing code (but returning Expression, if you see what I mean.)

Shout if you want a sample.

skyrising:
Ok, where do I hook the rewriter so it actually gets used?

leibnitz27:
Have a look at usage of HexLiteralTidier in miscKeyholeTransforms.

skyrising:
Ok, got it working for most cases, but it doesn't seem to get applied to field initializers

leibnitz27:
Fixed for field initialisers (though I'll go back and tweak a bit).

Now this is nicely split out it can be controlled better by a parameter too, yay!

I confess I'm a bit surprised by usages of pi/90 - radians divided by degrees feels odd, but I'm not here to legislate OTHER projects.... ;)

Thanks - great commit.

leibnitz27:
Added a few more ConstantNameTestXXX to test sets.

ConstantNameTest5 is particularly fun ;)

leibnitz27:
Added a few small improvements to this, https://github.com/leibnitz27/cfr_tests/blob/master/src_6/org/benf/cfr/tests/ConstantNameTest6a.java demonstrates behaviour.

PR #74, opened 2020-01-09
Fix Eclipse compilation issue

Marcono1234:
Eclipse failed compiling the changed line of CFR source code, possibly due to an Eclipse specific compiler bug since `javac` could compile the code just fine.

Eclipse compilation error

However, since `javac` can compile the code, I can understand if you decide to refuse merging this pull request.


leibnitz27:
I may have let slip that I'm not the world's biggest fan of Eclipse, but I'm not anti ;) :)

Thanks!

Issue #96, opened 2020-01-24
Type annotations placed incorrectly for scoped and qualified types


Marcono1234:
# CFR version
0.149-SNAPSHOT (commit da9725ba)

# Compiler
javac 11.0.5

# Description
Relates to #93
See [JLS 9.7.4](https://docs.oracle.com/javase/specs/jls/se13/html/jls-9.html#jls-9.7.4)

>A type annotation is _admissible_ if both of the following are true:
>- The simple name to which the annotation is closest is classified as a _TypeName_, not a _PackageName_.
>- If the simple name to which the annotation is closest is followed by "`.`" and another _TypeName_ - that is, the annotation appears as `@Foo T.U` - then `U` denotes an inner class of `T`.

Source:
```
import static java.lang.annotation.RetentionPolicy.RUNTIME;

import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.Target;

class TypeAnnotationPlacement {
    @Retention(RUNTIME)
    @Target({ElementType.TYPE_USE})
    @interface MyParam { }
    
    class List { }
    static class Nested1 {
        static class Nested2 { }
    }
    
    // https://docs.oracle.com/javase/specs/jls/se13/html/jls-9.html#jls-9.7.4
    Nested1.@MyParam Nested2 field;
    
    // Use return type here due to https://github.com/leibnitz27/cfr/issues/95
    java.util.@MyParam List<?> test() {
        return null;
    }
}
```

Decompiled output:
```
import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

class TypeAnnotationPlacement {
     // Invalid
    @MyParam TypeAnnotationPlacement.Nested1.Nested2 field;

    TypeAnnotationPlacement() {
    }

     // Invalid
    @MyParam java.util.List<?> test() {
        return null;
    }

    static class Nested1 {
        Nested1() {
        }

        static class Nested2 {
            Nested2() {
            }
        }
    }

    class List {
        List() {
        }
    }

    @Retention(value=RetentionPolicy.RUNTIME)
    @Target(value={ElementType.TYPE_USE})
    static @interface MyParam {
    }
}
```

PR #110, opened 2020-01-31
Annotation fixes

Marcono1234:
Some changes (and hopefully improvements) related to annotations.
I hope they are alright, let me know if you want anything changed.


src/org/benf/cfr/reader/bytecode/analysis/types/TypeConstants.java
```diff
@@ -1,22 +1,32 @@
...
+    JavaRefTypeInstance CHAR_SEQUENCE = JavaRefTypeInstance.createTypeConstantWithObjectSuper(charSequenceName);
+    JavaRefTypeInstance STRING = JavaRefTypeInstance.createTypeConstant(stringName, OBJECT, CHAR_SEQUENCE);
+    JavaRefTypeInstance CLASS = JavaRefTypeInstance.createTypeConstantWithObjectSuper(className);
+    JavaRefTypeInstance ITERABLE = JavaRefTypeInstance.createTypeConstantWithObjectSuper("java.lang.Iterable");
+    JavaRefTypeInstance CLOSEABLE = JavaRefTypeInstance.createTypeConstantWithObjectSuper("java.io.Closeable");
+    JavaRefTypeInstance SERIALIZABLE = JavaRefTypeInstance.createTypeConstantWithObjectSuper("java.io.Serializable");
+    JavaRefTypeInstance THROWABLE = JavaRefTypeInstance.createTypeConstantWithObjectSuper(throwableName);
+    JavaRefTypeInstance AUTO_CLOSEABLE = JavaRefTypeInstance.createTypeConstant("java.lang.AutoCloseable");
     JavaRefTypeInstance SUPPLIER = JavaRefTypeInstance.createTypeConstant("java.util.function.Supplier", "Object");
     JavaRefTypeInstance SCALA_SIGNATURE = JavaRefTypeInstance.createTypeConstant("scala.reflect.ScalaSignature", "Object");
```
Marcono1234: @leibnitz27, is it intended that `"Object"` is (and previously was) used as `shortName`?

Marcono1234: @leibnitz27, is it intended that `"Object"` is (and previously was) used as `shortName`?

leibnitz27:
Hey - thanks.  There's quite a lot here, so it might take me a little time to go through it.

Can you (sorry to be awkward with the different repo, reasons stated so I wont repeat) add some testcases to the lowest supporting version of java in cfr_tests?

(Hopefully you've run against a baseline generated from master already?)

leibnitz27:
(testcases possibly not covered as I tend to add them as I fix them, though I'm ok with having imperfect ones there to help spot odd regressions - the more the merrier)

Marcono1234:
I hadn't tested it against `cfr_tests` yet, but have tested against small examples and against a larger program I regularly decompile.
During testing I have found an issue I introduced so please **do not merge yet**. I will likely force-push corrected changes soon.

leibnitz27:
Ok.  

Is there any particular reason that you're looking at annotations in such detail?  I've certainly not focused on them, as they tend to be significantly irrelevant to the purpose of reading bytecode. (though I appreciate they could be used as a pattern to spot recompilation).

BTW - when adding tests, pls just add brand new ones - editing existing tests makes it harder to spot regressions, and the cost of duplicate code in tests is 0.  Thanks.

Marcono1234:
The reason why I am looking at annotations so much currently is that I am really just going through things which I consider tricky for a decompiler to get right. I had the idea of an ideal decompiler which is able to always produce code, which when compiled, produces functionally equivalent class files; at least for non-obfuscated code. However, now I noticed too that this is likely not possible or would complicate things extremely.
In general please let me know when I should stop reporting certain types of issues. And sorry for this flood of issues, feel free to close them or tag them with a label to indicate that they are not as important or not worth the effort.

Thanks for all your work on the decompiler!

leibnitz27:
Annotations are definitely an [expletive redacted] ;)

I suspect prototype and field annotations can be reproduced correctly, but I have a feeling it will be incredibly tricky to get all in-code ones correct.

I've also *massively* shot myself in the foot by forbidding usage of the localVariableTypeTable.  Ah well, I think it was the right decision overall, but it *really* makes some things hard.

If you're keeping a score card of how CFR's doing comparatively - I don't tend to try out other decompilers (except fernflower occasionally, due to idea integration ;) ) - drop me a mail ;)

leibnitz27:
But don't give up on this change if you're making progress, (or just pick a smaller scope) as you've identified, there's basically an infinity of stuff left out there. (combined sadfacehappyface emoji)

Lanchon:
> I've also massively shot myself in the foot by forbidding usage of the localVariableTypeTable.

it was the right choice!

Marcono1234:
@leibnitz27, ok forced pushed now. These changes are ready to merge now. I have tested against `cfr_tests` compiled with JDK 13 and against the large program I am regularly decompiling and could not find any newly introduced issues <sup>1</sup>.

I know that these are quite a lot of changes, but I tried to keep all of them grouped to one commit respectively. Therefore, I hope by going through the commits one by one (instead of looking at all changed files at once) the changes should make sense.

The class `DeclarationAnnotationHelper` contains some logic regarding non-admissible locations which is not completely used yet because it is currently not possible without rewriting the logic how type collisions are resolved and without being able to control how types are dumped.

----

<sup>1</sup>: I spotted one case where the `Main` class of `multijar2.jar` failed to load itself (according to the error comment), but I was not able to reproduce this again so maybe I made a mistake in the test setup that time.

leibnitz27:
You're right - this is a pretty hefty change.

Haven't looked at the code yet (may take a few days), but are you sure this is valid?

Original

```
public class AnnotationTestCode5 {
    public void test(AnnotationNested n) {
        @AnnotationCode2(name="a") AnnotationNested. @AnnotationCode2(name="b") Fred. @AnnotationCode2(name="c") Jim. @AnnotationCode2(name="d") Alan alan = n.getAlan();
        System.out.println(alan);
        AnnotationNested. @AnnotationCode2(name="b") Fred. @AnnotationCode2(name="c") Jim. @AnnotationCode2(name="d") Alan alan2 = n.getAlan();
        System.out.println(alan2);
        AnnotationNested.Fred. @AnnotationCode2(name="c") Jim. @AnnotationCode2(name="d") Alan alan3 = n.getAlan();
        System.out.println(alan3);
    }


}
```

Test difference

```
Checking AnnotationTestCode5
13c13
<         AnnotationNested. @AnnotationCode2(name="b") Fred. @AnnotationCode2(name="c") Jim. @AnnotationCode2(name="d") Alan alan2 = n.getAlan();
---
>         @AnnotationCode2(name="b") AnnotationNested.Fred. @AnnotationCode2(name="c") Jim. @AnnotationCode2(name="d") Alan alan2 = n.getAlan();
15c15
<         AnnotationNested.Fred. @AnnotationCode2(name="c") Jim. @AnnotationCode2(name="d") Alan alan3 = n.getAlan();
---
>         @AnnotationCode2(name="c") AnnotationNested.Fred.Jim. @AnnotationCode2(name="d") Alan alan3 = n.getAlan();

```

Marcono1234:
You are right, I must have missed that, probably because it could not load the outer class in my test setup so I did not pay enough attention to it.

However, the annotation is actually placed "correctly" here:
https://github.com/leibnitz27/cfr/blob/c0723f1e5801041db2710cb0717051e5c256ce8f/src/org/benf/cfr/reader/bytecode/analysis/types/JavaRefTypeInstance.java#L161

This tried to solve the issue that within an inner class it would emit all outer classes:
```
class A {
    class B {
        class C {
            A.B. @TypeUse C f;
            // With this change it becomes
            @TypeUse C f;
        }
    }
}
```

However the downsite I forget to consider is when the outer classes have to be used (as it is the case with your test case).
While this can be fixed in a hacky way by always dumping the outer class of the annotated type as type and then the actually annotated type only as name (= `B. @TypeUse C`), I would prefer not to do this.
Instead this could (and probably should) be tackled together with #96 by changing the dumping logic to allow passing annotations together with the type. I can try to take a look at that, however it appears to be quite some work and it will not become part of this pull request.

So I hope this change is fine for now.

src/org/benf/cfr/reader/bytecode/analysis/types/TypeConstants.java
Marcono1234: @leibnitz27, pinging you again.

leibnitz27:
Nah - there's a relatively easy way to check if the first annotation that is applied works against a shortened name.   It's a little ugly combining display with logic, but because annotations are done at display time (which I actually like, as it doesn't allow them to interfere with code patterns), it's kind of ok.

I've added a few more tests (Incl the purely evil AnnotationTestM2g) which I think might tickle bugs in this in the future.

FWIW - It really would be much easier if this was 2/3 different PRs - I know the urge to tidy stuff as you're passing is high (and yes, thanks, you have done it as different commits), but it makes the barrier to a PR higher. (especially if the tidying commit has depencies on earlier, more significant changes).

I'm still not 100% convinced that
```
@MyField 
public @MyParam int field;
```

is intrinsically better/more correct than 

```
public @MyField @MyParam int field;
```

I don't normally get particularly upset by styling where there isn't a strong JLS requirement, but the above nags a bit.

leibnitz27:
(I'll make that fix onto this)

leibnitz27:
(This is merged, but I merged it from a local with a few regression fixes in - sorry, bad gitiquette probably).

leibnitz27:
(Sigh, tagging, not requesting.  lol.)
Anyway, thanks ;)

leibnitz27:
Looks like this has fixed at least #113 as well ?

src/org/benf/cfr/reader/bytecode/analysis/types/TypeConstants.java
leibnitz27: Good point.

Marcono1234: So is it intended or not? And if not, should I create a pull request or do you want to change it?

Marcono1234:
Thanks for merging and sorry for this being containing so many commits, but I wanted to solve them all at once and because they depend on each other (at least partially) I would have had to wait for every pull request to be merged, though that would have been better for you and the tyding commits could have been separate.

Re: Annotation placement
```
@MyField 
public @MyParam int field;
```
In my opinion this better reflects the intention: `@MyField` annotates the field as a whole while `@MyParam` only annotates the type. I guess you could argue where to place annotations with declaration + `TYPE_USE` as target, but in my opinion it is more readable if the annotations are placed above. Especially with many annotations it is otherwise difficult to directly see the type of the field if it is shifted so far to the right due to all the annotations in front of it.

Re: Annotation placement and short name
Sadly your commit 5effaf2 does not fix the issue completely:
1. It now unnecessarily qualifies types using the outer again (in some cases):
    ```
    class Test {
        class A {
            class B { }
        }

        A. @TypeUse B f;
        // Becomes: Test.A. @TypeUse B f;
    }
    ```
1. The logic for finding the first annotated type seems faulty if the most inner type is annotated, it should probably be:
    ```
    Annotated getFirstWithEntries() {
        if (inner != null && entries.isEmpty()) return inner.getFirstWithEntries();
        return this;
    }
    ```
1. I feel like the logic in `dump(Dumper d)` should be moved to `dump(Dumper d, boolean hasDumpedType)` because currently the logic for finding the first annotated type is duplicated which makes it more error-prone if only one method is changed in the future. Also with the current code one could easily think that `dump(Dumper d)` can be refactored to directly dump `firstEntryType`, but that could discard the comment set for the outer type.

src/org/benf/cfr/reader/bytecode/analysis/types/TypeConstants.java
leibnitz27: Nah - leave it - I'll fix it when I get near a computer :)

Looks unintended, I believe I was after withObjectSuper but I'll need to run through the regression tests.

leibnitz27:
Makes sense - I'm going to leave  https://github.com/leibnitz27/cfr/commit/5effaf20af1bca791feaa6ba70b40daa0a55220d in for now, because it removes the regression in AnnotationTestCode5 - but you're of course right, it could be done better.

If you fancy having a go, shout, otherwise, I'll get there ;)

PR #112, opened 2020-02-01
Fix SUMMARY Sink implementations not receiving error information

Marcono1234:
Previously sinks would not receive errors logged using `SummaryDumper.notifyError(JavaTypeInstance, Method, String)`, this is now fixed.

This pull request introduces a new interface `MethodErrorCollector`.
I think ideally `Dumper` should not have the `addSummaryError(Method method, String s)` method in the first place because some dumper implementations (e.g. `ToStringDumper`) are not designed to handle errors. However, passing an error collector to all the places where the method is currently called would require a possible rather large rewrite so I choose to solve it by using the new interface by some dumper implementations only.

Please let me know what you think.


leibnitz27:
Indeed, I definitely don't want to thread additional stuff in all the places dumper is used.

That leads to a little ugliness, but pragmatically, it's much less ugly than the alternative.   If a dumper is considered to collect emissable information then this definition infact doesn't conflict at all....

PR #120, opened 2020-02-28
Give more info about identifiers to Dumper

Runemoro:
We'd like to use CFR as an alternative to Procyon in [Enigma](https://github.com/FabricMC/Enigma), a tool for remapping obfuscated Java code. See [this PR](https://github.com/FabricMC/Enigma/pull/192). To do this, we need to build a mapping between code regions and the entries they refer to. This PR extends the `Dumper` class to add:
 - `parameterName(String name, MethodPrototype method, int index, boolean defines)`
 - A `defines` parameter to the `dump(JavaTypeInstance javaTypeInstance)` method, which is set to true when the type instance is the class name in a class definition
 - A `field` parameter to the `fieldName` method, since we need the field type too (obfuscation can rename several fields to the same name, so just the name and owner isn't enough).

### DIFF
--- a/src/org/benf/cfr/reader/bytecode/analysis/parse/lvalue/AbstractFieldVariable.java
+++ b/src/org/benf/cfr/reader/bytecode/analysis/parse/lvalue/AbstractFieldVariable.java
@@ -114,6 +114,10 @@ public abstract class AbstractFieldVariable extends AbstractLValue {
         return classFileField;
     }
 
+    public Field getField() {
+        return classFileField == null ? null : classFileField.getField();
+    }
+
     @Override
     public SSAIdentifiers<LValue> collectVariableMutation(SSAIdentifierFactory<LValue, ?> ssaIdentifierFactory) {
         //noinspection unchecked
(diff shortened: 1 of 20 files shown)


leibnitz27:
This implies you're implementing dumper yourself - is this the case?  I make no guarantees about this interface, indeed, it changes *very* frequently.

leibnitz27:
Yeah, I just had a look at your other PR.

I would *strongly* recommend against doing this - you will require significant work every time you change versions of CFR.

https://github.com/leibnitz27/cfr_client

describes the stable interfaces, which I go to a lot of effort to *avoid* breaking backwards compatibility with.

leibnitz27:
ping @Runemoro

Runemoro:
So the proper solution would be to extend the token sink with class tokens, and add type + owner information for field and method tokens, right?

I've been busy this week and didn't have time to work on this. I might have some time during the weekend.

leibnitz27:
Yup.

One thing though - it might be worth discussing this in more detail first, as I don't want to expose (eg) JavaTypeInstance.   I've deliberately only made guarantees about stuff under org.benf.cfr.api.

i.e. could expose a string type, but that loses generic info.  It's a pain, and I may have to end up making a stable type class.

Why can't I just use java's?  Because I'm locked to java6 (deliberately, heh), and I don't want to have to use java XXX just to expose its features.

leibnitz27:
Going to close this - as described, dumper is not a stable interface, and implementing this will cause dependency issues,

Issue #104, opened 2020-01-26
`_` is not considered a keyword


Marcono1234:
# CFR version
0.149-SNAPSHOT (commit da9725ba)

# Compiler
javac 11.0.5

# Description
`_` (underscore) is not considered a keyword even though it is one since Java 9 ([JLS 3.9](https://docs.oracle.com/javase/specs/jls/se9/html/jls-3.html#jls-3.9)).

Renaming types, methods or fields should probably only be done when explicitly enabled. However local variable names (when compiled with `-g:vars`) should possibly always be renamed. Or both should be done based on the class file version?

[Here](https://github.com/leibnitz27/cfr/files/4113554/Test.class.txt) (remove suffix `.txt`) is a class file which claims to be Java 11 (major 55, minor 0) but uses `_` for a field and a local variable. Neither of them are renamed regardless of the `--renameillegalidents` value.

leibnitz27:
FWIW, this is waaaaay more fun (pain) than you'd think.

In order to handle this nicely, I need an initial pass through all classes that are being processed (which I've resisted thus far) - as it's possible to have a java6 class file which has a public field '_' (not strictly bad), which is LATER referenced by a java 9+ class in the same (batch/jar) (which, yeah, shouldn't happen, but remapping needs to be consistent.).

Fun ;)

Marcono1234:
Or would it make sense to allow specifying the output Java version?

Marcono1234:
Though that also applies to already considered keywords then (see `Keywords.isAKeyword`):
- `strictfp` became a keyword in Java 2
- `assert` became a keyword in Java 4
- `enum` became a keyword in Java 5

leibnitz27:
True.    If I get a general solution working for this I may change that. ;)

Annoyingly, adding a pass (checking only the first N bytes for version) is still horribly slow, adding a totally unacceptable 20s to my mahoosive test jar.

The case where a old field is used in a new version (so is legal when the old is considered on its own, but not in the context of the new) is pretty unlikely, so I may end up just warning about a symbol that is 'fixed' in the 9+ but not in the source. (unless a override version is set explicitly).

This is definitely an entertaining one - I thought it'd be 30s on the face of it, but .... can of worms ;)

Marcono1234:
> unless a override version is set explicitly

Is that an option / feature already (sorry if I overlooked that) or are you planning to add that?

leibnitz27:
[...]

PR #152, opened 2020-05-02
Fix identifiers containing ignorable chars not being renamed

Marcono1234:
Ignorable chars (Character.isIdentifierIgnorable) are ignored as part of identifiers (cannot be leading, but can be in between and trailing) when compiling. They can therefore normally not appear in a class file unless it was modified. Since they cannot be represented in source (and could easily cause identifier collisions), identifiers containing them are now considered invalid.

This was only added to the JLS for Java 10 (see [JDK-6214519](https://bugs.openjdk.java.net/browse/JDK-6214519)), however the behavior also exists in earlier versions.

Demonstration of the issue:
1. Download [IdentifierIgnorable.class.txt](https://github.com/leibnitz27/cfr/files/4566729/IdentifierIgnorable.class.txt) (remove `.txt`, that is only to make GitHub happy)
2. Decompile with `--rename true`

### DIFF
--- a/src/org/benf/cfr/reader/util/output/IllegalIdentifierReplacement.java
+++ b/src/org/benf/cfr/reader/util/output/IllegalIdentifierReplacement.java
@@ -42,7 +42,10 @@ public class IllegalIdentifierReplacement implements IllegalIdentifierDump {
             return true;
         }
         for (int x=1;x<chars.length;++x) {
-            if (!Character.isJavaIdentifierPart(chars[x])) return true;
+            char c = chars[x];
+            // If identifier contains ignorable char, it cannot be represented in source
+            // because e.g. `a\u0008` is effectively `a` when re-compiled
+            if (!Character.isJavaIdentifierPart(c) || Character.isIdentifierIgnorable(c)) return true;
         }
         return false;
     }


Lanchon:
i thought this ident ignorable thing was never implemented.

Marcono1234:
Looks like this was fixed for javac in Java 8: [JDK-7144981](https://bugs.openjdk.java.net/browse/JDK-7144981)

Lanchon:
ok i didnt look into ur link in detail. but if it was "fixed in javac" only and not in the JVM it'd mean that id ignorable chars apply to the JLS only. bytecode itself would be stripped of those chars and the JVM would (typically) never find them. this would imlply that changes to the JVM are not needed, and that changes to CFR are not needed either.

cfr should only be changed if the JVM ignores the chars. and then, it should probably keep them, but the compare operations should ignore them. they should be carried to source where theyd eventually be ignored by javac.

Marcono1234:
Yes, the are only ignored during compilation so they would normally not appear in class files, but in case they appear there anyways, the JVM apparently treats them like every other character.

The reason why I think CFR should consider them is that CFR emits source code, and the problem with ignorable chars is that they easily allow identifier collisions (as shown in the attached class file linked above). E.g. the following cannot be compiled:
```
class Test {
    int ab;
    int a\u0008b;
}
```
Therefore I think it should be CFR's task to rename them (if the option is set).

Lanchon:
IMHO CFR shouldnt rename. a byecode renaming pass (different tool) should be invoked first if you want your bytecode renamed.

Marcono1234:
That is a different discussion I guess. The `--rename` aggregate option and the more specific options (relevent here is `--renameillegalidents`) already exist and in my opinion they should therefore consider this case.

Lanchon:
yes i agree, i was sort of changing the subject without first stating that i saw your point, as i should have. im very tired right now, sorry.

ive been following these ident rename things for a while and im meaning to propose an alternative solution to Lee for months now but i hadnt have the time :(

leibnitz27:
Hey - looks good - sorry, been otherwise distracted over weekend.    Will have a play tonight - thanks!

Marcono1234:
No problem (to both of you).
Should this also depend on the Java version of the class file similar to #104? `Character.isIdentifierIgnorable` was added in 1.1, but the functionality was apparently added in Java 8 ([JDK-7144981](https://bugs.openjdk.java.net/browse/JDK-7144981)) and was included in the JLS in Java 10 ([JDK-6214519](https://bugs.openjdk.java.net/browse/JDK-6214519)). Though not sure how compilers other than `javac` handle it.

leibnitz27:
I'd love for that to work (class file version dependency), but that makes things inordinately more painful, when a java 6 class file (not illegal) references a java 8 class file (illegal).

I don't (and would prefer not to ) impose an analysis ordering (that feels brittle) - and as established in #104 , an initial pass is prohibitively expensive. 

The other issue is that CFR totally cheats here, with regards to aliasing illegal identifers to good ones - it simply generates a lookup for the lifetime of the decompilation - this means we don't have to actually check with b.class if its identifier has been renamed, we simply say 'oh, invalid, what does that rename to'.

I don't currently have a good solution for this - I know the above is a cheat, but keeping complete track of relationships adds a whole messy bookkeeping I'd rather not do....

PR #156, opened 2020-05-10
Add GitHub Maven build workflow

Marcono1234:
This adds a workflow to build the project on GitHub for every push and pull request (regardless of branch).

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,19 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: Java CI with Maven
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Build with Maven
+      run: mvn --batch-mode --update-snapshots -Dgpg.skip=true clean verify
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -15,5 +15,11 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
+    - name: Cache local Maven repository
+      uses: actions/cache@v1
+      with:
+        path: ~/.m2
+        key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+        restore-keys: ${{ runner.os }}-m2
     - name: Build with Maven
       run: mvn --batch-mode --update-snapshots -Dgpg.skip=true clean verify


leibnitz27:
This looks like it uses 1.8 - I'm (yes, I know, I'm perverse) keen to keep this as 1.6.

Is that possible with the tools provided?

Marcono1234:
It uses JDK 1.8 to run Maven, but after all the `pom.xml` defines the `--source` and `--target` arguments for `javac`, which for this project is 1.6. This could cause issues when you accidentially use APIs which were added after 1.6 (also described in [this comment by Brian Goetz](https://stackoverflow.com/questions/54843571/java-11-what-is-the-difference-between-javac-release-and-source-and-target#comment96462400_54843571)).

While Zulu (which is used by `actions/setup-java`) still provides JDK 1.6, there [won't be any updates](https://www.azul.com/products/azul-support-roadmap/).
Additionally using JDK 1.6 with Maven is not possible either, because [Maven 3.3+ requires JDK >= 1.7](https://maven.apache.org/download.cgi). So changing the JDK version used by this GitHub workflow is not an option.
How do you build the project locally? Do you actually use JDK 1.6 and Maven <= 3.3?

For now this workflow hopefully at least adds some value.

Ideally this project would use [Maven toolchains](https://maven.apache.org/guides/mini/guide-using-toolchains.html) to allow specifying JDK >= 9 when running Maven and then using the `javac` `--release` option to make sure it compiles against the correct API.

leibnitz27:
So as crazy as it seems, one reason I haven't tried to go for too much server side build automation (other than the fact I never really expected people to be taking head of master much, lol) is because I expect my weird constraints to bang up against reality....

That being said, if this is accurately using the --source and --target then great.

A follow up to this (which I'll have to figure out) is that currently I pull displayed version information out of the jar (just ` getPackage().getImplementationVersion() ` ).     <-- yeah, this isn't ideal, esp when someone hugejars it, tbh I naively wasn't really paying attention to that, but it's annoying.

Building more frequently like this would definitely make sense to try to compile the 

* release
or 
* for snapshot, the commit hash

into the jar.

My mvn fu isn't particularly strong, so if you have any ideas, would be keen to include them in this PR.

Lanchon:
the source/target javac options change javac behavior only. you can still use newer java APIs unless you set the boot classpath too. but its better than nothing.

FYI, the value 1.6 for these options has already been deprecated in newer javacs, but is still honored in 1.8 javac.

leibnitz27:
Indeed - hence why the tests repository also explicitly uses multiple versions of java, from 1.6 -> 14+.

I had a look at trying to bake that behaviour into integration tests in this repository, but, well, shall we say it's a pretty unusual use-case, so I let discretion be the better part of valour ;)

Marcono1234:
@Lanchon
> FYI, the value 1.6 for these options has already been deprecated in newer javacs, but is still honored in 1.8 javac.

It is also still supported by JDK 11 (and maybe even 12?), but already logs a warning there:
> source value 6 is obsolete and will be removed in a future release

@leibnitz27, I have created a separate pull request ( #157) for providing build information and not relying on `Package.getImplementationVersion()` anymore.

Would you mind merging this branch then?

leibnitz27:
Might have to get a bit creative when 1.6 disappears into even more distant history, but that's for another day.  

Thanks!

PR #157, opened 2020-05-10
Provide more build information to CFR, don't use getImplementationVersion()

Marcono1234:
The version information now includes the commit hash and whether the working tree was dirty when the project was built.
It doesn't use `Package.getImplementationVersion()` anymore and should therefore continue to work even if CFR is packaged in a fat / uber jar.

To accomplish this, it uses
- https://github.com/git-commit-id/git-commit-id-maven-plugin
To get the Git commit hash and to determine whether the working tree is dirty
- https://www.mojohaus.org/templating-maven-plugin/
To populate the `CfrVersionInfo.java` with the project version and Git information

The result when running `java -jar cfr-0.150-SNAPSHOT.jar` is (for example):
```
CFR 0.150-SNAPSHOT (64ce6e4)
```
Respectively if the working tree is dirty:
```
CFR 0.150-SNAPSHOT (64ce6e4-dirty)
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,7 @@
         <profile>
             <id>jdk9plus</id>
             <activation>
-                <jdk>[1.9,)</jdk>
+                <jdk>[9,)</jdk>
             </activation>
             <build>
                 <plugins>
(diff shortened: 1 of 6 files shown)


leibnitz27:
Love it!

leibnitz27:
FWIW - I'm torn about leaving the commit-id in the non-snapshot version - it's massively useful in snapshots, but not so much.   On the other hand, more info > less info ;)

PR #165, opened 2020-05-13
Fix  #160: Add GitHub issue templates

### LINKED ISSUE #160: Add issue templates
What do you think about adding issue templates (see [GitHub help article](https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository))?

They would allow defining an issue description which could then indicate to the issue creator which information they have to provide, e.g. CFR version and compiler version.
Additionally issue templates can automatically add labels, so you could for example create two templates, one for bugs and one for feature requests. However, users are by default not restricted to these and can still create a normal issue if they think none of the templates matches their issue.

Would you like me to create an example pull request for it?

Marcono1234:
Please let me know if you want anything to be changed, or feel free to change it yourself.

@Col-E and @GraxCode, since you have written multiple issues here as well, what do you think about these templates?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,26 @@
+---
+name: Bug report
+about: Report a bug with CFR.
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+# CFR version
+<!-- Specify the CFR version you are using, example:  "CFR 0.150-SNAPSHOT (64ce6e4)" -->
+<!-- This information can be obtained by running `java -jar cfr.jar` -->
+
+
+# Compiler
+<!-- Specify the compiler and the version you are using -->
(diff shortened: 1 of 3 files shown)


GraxCode:
Pretty nice

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,26 @@
...
+<!-- For `javac` paste the output of `javac -version` -->
+
+
+# Description
+<!-- Describe the bug, include expected vs. actual behavior if necessary -->
+
+
+# Example
+<!-- Provide source code to compile or attach `.class` files or `.jar`s (append the extension `.txt` or similar to make GitHub allow the upload) -->
+<!-- IMPORTANT: Only provide code or class files which you are legally allowed to share here -->
```
leibnitz27, line 26: Thanks for adding the disclaimer!

Issue #136, opened 2020-04-09
[Feature/API] More verbose error sink details


Col-E:
In my project I'm using the provided `OutputSinkFactory` for logging decompilation errors. However this relies on the exception message being non-null, which is not always the case. For example [`Driver.doClass(...)` has `ed.noteException(path, null, e);`](https://github.com/leibnitz27/cfr/blob/8f7478d99fb44687c28547269afabc41eb062fc2/src/org/benf/cfr/reader/Driver.java#L103)

![image](https://user-images.githubusercontent.com/21371686/78909226-334f5c00-7a51-11ea-8ece-1725e3f6abfd.png)

It would be very useful to also be provided the actual `Exception` instance

PR #179, opened 2020-06-04
Missing quotes

GraxCode:


Marcono1234:
Oh, I always thought this was intended to prevent recompiling it and make you aware that there was a decompilation failure.
Would it be possible to highlight this issue better then (e.g. add a comment as well), otherwise it just looks like a normal exception and depending on what you decompile (e.g. you decompile a decompiler) you would not even notice it.

Col-E:
> Oh, I always thought this was intended to prevent recompiling it and make you aware that there was a decompilation failure.

From an API standpoint, I would prefer this in combination with reporting any decompile failures/exceptions in the logging sink.  _(Plugging #136 )_

leibnitz27:
Yeah - good points.  Maybe a 'remove me' which DOES fail to compile next to it.

leibnitz27:
(and #136 ;) )

Marcono1234:
It also has its advantages that it emits compilable code because it probably makes navigation in that code using an IDE easier. Though fixing issue 136 and maybe (?) a comment would be useful nonetheless.

GraxCode:
I think a //FIXME: would make it easier to highlight decompilation errors, and IMO producing runnable code (even if it throws an exception) is better than producing invalid code

PR #181, opened 2020-06-04
Properly handle abnormal booleans

GraxCode:


leibnitz27:
The reason for 'BADBOOL 3' was because this is _usually_ found when CFR has incorrectly determined that something is boolean when it's actually an integer - i.e. it's a last gasp "clearly made the wrong typing decision, but I don't want you to lose that the value really was a 3".

This change discards information - have you got an example where it's happening?

GraxCode:
[boolean-minus-one.jar.txt](https://github.com/leibnitz27/cfr/files/4732763/boolean-minus-one.jar.txt)
Boolean with value -1 is `true`, also tested with 2,3 and other positive integers.

GraxCode:
You could also interpret booleans that are not 0 or 1 as "X != 0", if you want to keep the information about which number it is

leibnitz27:
I like that - let's turn badbool into != 0.  Do you want to do it?

Issue #149, opened 2020-04-26
Decompile handle LDCs


GraxCode:
[NewLDCTypes.class.txt](https://github.com/leibnitz27/cfr/files/4536629/NewLDCTypes.class.txt)
Currently, CFR fails to decompile Handle ldcs. 
```
 /*
         * This method has failed to decompile.  When submitting a bug report, please provide this stack trace, and (if you hold appropriate legal rights) the relevant class file.
         * org.benf.cfr.reader.util.ConfusedCFRException: Expecting a ConstantPoolEntryLiteral or ConstantPoolEntryDynamicInfo
         * org.benf.cfr.reader.bytecode.opcode.OperationFactoryLDC.getStackType(OperationFactoryLDC.java:37)
         * org.benf.cfr.reader.bytecode.opcode.OperationFactoryLDC.getStackDelta(OperationFactoryLDC.java:18)
         * org.benf.cfr.reader.bytecode.opcode.JVMInstr.getStackDelta(JVMInstr.java:315)
         * org.benf.cfr.reader.bytecode.analysis.opgraph.Op02WithProcessedDataAndRefs.populateStackInfo(Op02WithProcessedDataAndRefs.java:195)
         * org.benf.cfr.reader.bytecode.analysis.opgraph.Op02WithProcessedDataAndRefs.populateStackInfo(Op02WithProcessedDataAndRefs.java:1542)
         * org.benf.cfr.reader.bytecode.CodeAnalyser.getAnalysisInner(CodeAnalyser.java:393)
         * org.benf.cfr.reader.bytecode.CodeAnalyser.getAnalysisOrWrapFail(CodeAnalyser.java:251)
         * org.benf.cfr.reader.bytecode.CodeAnalyser.getAnalysis(CodeAnalyser.java:185)
         * org.benf.cfr.reader.entities.attributes.AttributeCode.analyse(AttributeCode.java:94)
         * org.benf.cfr.reader.entities.Method.analyse(Method.java:521)
         * org.benf.cfr.reader.entities.ClassFile.analyseMid(ClassFile.java:1035)
         * org.benf.cfr.reader.entities.ClassFile.analyseTop(ClassFile.java:922)
         * org.benf.cfr.reader.Driver.doJarVersionTypes(Driver.java:252)
         * org.benf.cfr.reader.Driver.doJar(Driver.java:134)
         * org.benf.cfr.reader.CfrDriverImpl.analyse(CfrDriverImpl.java:65)
[...]

PR #183, opened 2020-06-05
More literals & partial support for MethodHandle LDCs

GraxCode:


GraxCode:
I wanted to finish the method handle and type expressions, but realized I am too unfamiliar with the CFR source code 😬

leibnitz27:
Can this be tested against https://github.com/leibnitz27/cfr/issues/149 ?

Can you do a pr with some tests to https://github.com/leibnitz27/cfr_test ?

If you have hardcoded tests (i.e. it's not easy to get java to create), then just stick the bytecode under 'hardcoded'.

GraxCode:
This does not fix #149, as it only comments the handle, so the rest of the code is readable

GraxCode:
[NewLDCTypes.class.txt](https://github.com/leibnitz27/cfr/files/4735991/NewLDCTypes.class.txt)
[MethodTypeFun.class.txt](https://github.com/leibnitz27/cfr/files/4735992/MethodTypeFun.class.txt)
I am not going to finish the MethodHandle implementation, as I'm probably using the Expression class wrongly. Don't know what to do with InferredJavaType...

leibnitz27:
I'll merge this as I don't think it leaves anything any worse off.

When doing a PR like this, can you try to keep separate changesets in different PRs?  The Literals are a very different thing - it would have been handy to have separate PRs.

leibnitz27:
My mistake - looks like you've taken MethodType about as far as it can go already - using `MethodType.fromMethodDescriptorString` is a nice touch ;)

src/org/benf/cfr/reader/bytecode/analysis/opgraph/Op02WithProcessedDataAndRefs.java
```diff
@@ -1473,13 +1473,36 @@ private Statement createStatement(final Method method, VariableFactory variableF
...
+      if (cpe instanceof ConstantPoolEntryMethodHandle) {
+        return constructMethodHandleExpression((ConstantPoolEntryMethodHandle) cpe);
+      }
+      if (cpe instanceof ConstantPoolEntryMethodType) {
+        return constructMethodTypeExpression((ConstantPoolEntryMethodType) cpe);
+      }
+      throw new ConfusedCFRException("Constant pool entry is neither literal, dynamic literal, method handle or method type.");
+    }
+
+    private static final InferredJavaType OBJECT_TYPE = new InferredJavaType(RawJavaType.NULL, InferredJavaType.Source.EXPRESSION);
```
leibnitz27, line 1491: The purpose of the InferredJavaType (as distinct from the JavaTypeInstance) is to allow types to be changed after we have more information about them - i.e an inferred java type might be assumed to be bool, but used in a context where it's definitely int - rather than have massive action/subscriber chains, this allows the change to be visible in all sites where it's used.

Because of this, a static inferredjavatype is probably a bad idea - if this was used in one context, it would affect others.

There's a "locked" argument to IJT constructor which forbids this, but it's probably safer here to simply not use a static - will fix.

leibnitz27: Also we know this is returning a MethodType, so we should just use that. (done in https://github.com/leibnitz27/cfr/commit/1d0be9c9743dbb7e199d1f1ae1e45437881a4cb8 )

leibnitz27:
Hmm- actually I'm not really sure HandleExpression is required.

Will follow up on #149

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/HandleExpression.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public HandleExpression(ConstantPoolEntryMethodHandle handle) {
+      super(new InferredJavaType(handle.getDefaultType(), InferredJavaType.Source.FUNCTION, true));
+      this.handle = handle;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null) return false;
```
Marcono1234, line 23: Note that a `null` check here is not needed since the `! instanceof` below covers that.
Though it does not matter that much.

Issue #73, opened 2020-01-03
Debugfriendly output possible?


jdelker:
I would like to use the decompiled java code for debugging purposes (e.g. tracing code execution in Netbeans), which basically requires two things:
- output of original line numbers
- fitting (stretching/combining) code lines to match actual lines in output file

[Procyon](https://bitbucket.org/mstrobel/procyon/wiki/Java%20Decompiler) does actually a pretty good job producing such output, but unfortunately fails in other areas.
Thus, I would love to see this being possible in CFR.

@leibnitz27: 
- Is this something you would think is feasible with CFR in a foreseeable future?
- Considering the current code and data structures, would you say this is something rather easily/hard thing to implement?

jdelker:
This somewhat includes issue #72 - assuming the author's intention was also to use line numbers for debugging.

Lanchon:
stretching/combining lines would detract a lot from readability, just when you need it most. plus the original jar would need to include line numbers, which in non-open source bytecode is rare. for these reasons, IMHO this feature is not a good idea.

from a usability perspective, a better spin at it would be having CFR reoutput the class files with added (or replaced) debugging info matching its output.

unfortunately this is very probably off the table because:
- in general, it would require completely new functionality to output class files, meaning a lot of effort for low returns and major feature creep for a decompiler.
- and specifically for CFR, i believe this decompiler uses its own class file reader instead of a library like ASM, meaning that writing class files would probably require rolling its own class writer library too, and thus would be a huge effort for this project in particular.

but there is a compromise:

assuming CFR's infrastructure allows establishing a relationship between bytecode addesses and line numbers of generated source (which may very well not be true), CFR could output maps of line numbers to bytecode addresses for each method body in some suitable format (standard, easy to parse). then someone else can easily create a tool based on ASM that can rewrite classes/jars and insert the debugging info from these mapping.

this is doable but requires new infrastructure in CFR to output the maps, and thus it is not completely ideal.

so here is an improvement:

maybe the least intrusive way would be for the author to define a custom parseable comment structure added at the end of some source lines that links this line (and next lines until another such comment) to a particular bytecode offset in the current method. this would require the least effort (no new infrastructure), and would still allow the creation of a bytecode rewriting tool.

if considering this solution, another custom parseable comment structure would be needed at the beginning of each method, linking the following bytecode offset comments to a specific bytecode method. this would accomplish two things:
[...]

PR #187, opened 2020-06-17
Made codeAttribute public

KOLANICH:
Needed to get access to `lineNumberTable` when using the decompiler as a lib

### DIFF
--- a/src/org/benf/cfr/reader/entities/Method.java
+++ b/src/org/benf/cfr/reader/entities/Method.java
@@ -90,7 +90,7 @@ public class Method implements KnowsRawSize, TypeUsageCollectable {
     private final AttributeMap attributes;
     private MethodConstructor isConstructor;
     private final int descriptorIndex;
-    private final AttributeCode codeAttribute;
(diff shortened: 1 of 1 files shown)


leibnitz27:
There are multiple issues here.

Relying on something which is not in the api means that you're effectively locking yourself to one/X version(s) of CFR.   While these things don't change very often, I make absolutely no guarantee about them.  (yes, I've been sloppy about marking things as package etc.   That's a fair criticism.  It's mainly because java's lack of hierachical permissions on packages absolutely sucks, but I don't want CFR to move to 9 to get modules).

But the bigger one is that LineNumberTable is something I absolutely will not use for analysis, and should not be trusted.

The line number table which is provided in the class file absolutely (I'll make this guarantee now) will not match the output of CFR.   

Providing it as a 'blessed' output makes that assumption much harder to explain.

Please have a look at https://github.com/leibnitz27/cfr/issues/73 - this is the only context I intend to provide line numbers downstream (and yes I discuss stretching squashing etc).

PR #196, opened 2020-06-28
Handle unusual static initializer access

GraxCode:
This class is completely runnable, even though the <clinit> method has all possible access flags except static.
[ClinitWithAccess.class.txt](https://github.com/leibnitz27/cfr/files/4842103/ClinitWithAccess.class.txt)
This is possible due to:
https://github.com/Dxh7788/openjdk/blob/986a51503a7a4a938c5ef31cd999787eac2d4759/openjdk/hotspot/src/share/vm/classfile/classFileParser.cpp#L2284-L2296

### DIFF
--- a/src/org/benf/cfr/reader/entities/Method.java
+++ b/src/org/benf/cfr/reader/entities/Method.java
@@ -451,6 +451,14 @@ public class Method implements KnowsRawSize, TypeUsageCollectable {
         boolean usesAdmissibleType = !annotationsInfo.requiresNonAdmissibleType();
         dumpMethodAnnotations(d, annotationsInfo.getDeclarationAnnotations(usesAdmissibleType));
 
+        if (isConstructor == MethodConstructor.STATIC_CONSTRUCTOR) {
+            // JVM Spec 2nd ed., chapter 4.6: All access flags except static for class initializers are ignored
+            // Pre java 7 class files even allow static initializers to be non-static
+            // See classFileParser.cpp#parse_method
+            d.keyword(AccessFlagMethod.ACC_STATIC.toString());
+            return;
+        }
+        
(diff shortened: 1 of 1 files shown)


leibnitz27:
Cool spot - but have you run this on the command line?  It's only going to affect methods that are marked safe for dumping ( so doesn't alter the output I see, unless I specify showing bridge methods - see below ).

Rather than having this in the dumper, it would be better to put it in the constructor for method, just before 

```
+++ if it's clinit, &= static|strict
        this.attributes = new AttributeMap(tmpAttributes);
```

```
java org.benf.cfr.reader.Main c:\Downloads\ClinitWithAccess.class
/*
 * Decompiled with CFR 0.151-SNAPSHOT (b266dda).
 */
public class ClinitWithAccess {
    public static void main(String[] args) {
    }
}

java org.benf.cfr.reader.Main c:\Downloads\ClinitWithAccess.class --hidebridgemethods false
/*
 * Decompiled with CFR 0.151-SNAPSHOT (b266dda).
 */
public class ClinitWithAccess {
    static {
        System.out.println("Okay");
    }

    public static void main(String[] args) {
    }
}
```

leibnitz27:
if you want to do that, shout (or just resubmit ;) ) otherwise I'll stick it in this evening.

Cheers!

GraxCode:
Github doesn't want to work today for me, you can do it. I wouldv'e added it here: https://github.com/leibnitz27/cfr/blob/b266ddaf3ea3acfafa8720346461a4a9ed9785ac/src/org/benf/cfr/reader/entities/Method.java#L129-L131
Clear the access flags and add static should solve that problem.

leibnitz27:
Done as described in https://github.com/leibnitz27/cfr/commit/94963b3c1ad140518f3f197f9f7b18e494d66fc2

PR #200, opened 2020-07-17
Use popular way of transposing var names that are reserved (z instead of s)

GraxCode:


Col-E:
So `class_` --> `clazz` and `boolean` --> `boolean_`  since it does not follow the pattern

leibnitz27:
While clazz is very common, I'm not sure I buy azzert, zynchronized, zuper, conzt, ztrictfp zwitch etc.

I think clazz is a one off here - rather than a general substitution perhaps a lookup would be better?

Col-E:
While we're here... [shouldn't `var`](https://github.com/leibnitz27/cfr/blob/375369a6cbcf3dc6d2678cf99b398567c0916015/src/org/benf/cfr/reader/bytecode/analysis/variables/Keywords.java) be added to the keyword list?

GraxCode:
so i've seen interfaze, clazz, zuper, but you're right, i've never heard of zynchronized lol (but how would that end up in variable anyway, everyone would use "synched"). Would still be cleaner (and funnier) to avoid "_". Maybe we should consider a Map for keyword variable names?

Lanchon:
merely a style issue, but i hate clazz. has nothing to do with java's usual naming practices. i use theClass.

Marcono1234:
> While we're here... [shouldn't `var`](https://github.com/leibnitz27/cfr/blob/375369a6cbcf3dc6d2678cf99b398567c0916015/src/org/benf/cfr/reader/bytecode/analysis/variables/Keywords.java) be added to the keyword list?

@Col-E, `var` as well as `yield` are not keywords but [restricted identifiers](https://docs.oracle.com/javase/specs/jls/se14/html/jls-3.html#jls-3.9) and therefore only forbidden in some contexts.
However, would you mind creating a separate issue for this since it will either get lost here or cause this pull request to drift off-topic.

GraxCode:
I created a draft of all replacements. Feel free to change / suggest anything if you don't like it.

GraxCode:
I forgot to add that i also added a check for "_" as name in java 9+ name (which I probably should remove, as there is a better solution, considering #104). I thing the best way would be to make a Keyword class with those fields:
```java
String name;
String replacement;
ClassFileVersion since;
```

src/org/benf/cfr/reader/bytecode/analysis/variables/Keywords.java
```diff
@@ -19,8 +19,30 @@
...
+        "bool", "_do", "_if", "privt", "that",
+        "_break", "doubleVal", "impl", "protctd", "_throw",
+        "byteVal", "elze", "imprt", "pblc", "throwz",
+        "caze", "enumVal", "instanzeof", "ret", "tranzient",
+        "_catch", "extendz", "intVal", "shortVal", "_try",
+        "_char", "_final", "interfaze", "_static", "_void",
+        "clazz", "finaly", "longVal", "strict", "volatil",
+        "constant", "floatVal", "_native", "_super", "_while",
+        "tru", "fals", "nullVal" 
+    );
```
leibnitz27, line 39: I get where you're going with this, but it's a really bad idea to assume that people pay enough attention to detail to keep two unrelated lists in synch.  
Sorry, zynch.

;)

Better to use a map which is statically constructed - I'll accept this change then tweak it to do that.

leibnitz27: Though I'm not sure I'm with you on half of these.   Will ponder..............

GraxCode: As i said, it's only a draft of how it could look like ;) Change what you want

PR #204, opened 2020-08-10
Attempt to remove try blocks without throwing/printable code

cookiedragon234:
Currently bytecode like the following:
```
A:
GOTO C
B:
// do something with the exception
C:
Try A->B Handled by B
```
Will result in an output like the following:
```Java
try {
} catch (Throwable t) {
	// do something with the exception
}
```

This is especially obvious in radon obfuscated applications where random try catches are added to single instructions like `goto`.

For now I have created a check where if try blocks contain no printable statements, and no statements that can throw an exception, the try block is removed.

This could be taken a step further, and all try blocks that are unable to throw an exception could be removed, however this would result in inaccurate output on valid javac classes. Let me know if you want this implemented and I can modify the PR.

Example output before:
```Java
public class hhgg {
    private static /* bridge */ /* synthetic */ void ggghhg() {
        boolean bl;
        try {
        }
        catch (hghghhhg hghghhhg2) {
            hghghhhg2.ghgggghh();
            throw hghghhhg2;
        }
        try {
        }
        catch (hghghhhg hghghhhg3) {
            hghghhhg3.ghgggghh();
            throw hghghhhg3;
        }
        try {
        }
        catch (hghghhhg hghghhhg4) {
            hghghhhg4.ghgggghh();
            throw hghghhhg4;
        }
        try {
        }
        catch (hghghhhg hghghhhg5) {
            hghghhhg5.ghgggghh();
            throw hghghhhg5;
        }
        try {
        }
        catch (hghghhhg hghghhhg6) {
            hghghhhg6.ghgggghh();
            throw hghghhhg6;
        }
        try {
            bl = ggghghg;
        }
        catch (hghghhhg hghghhhg7) {
            hghghhhg7.ghgggghh();
            throw hghghhhg7;
        }
        if (!bl) {
            String string;
            String[] stringArray;
            String[] stringArray2;
            int n;
            try {
                n = 3;
            }
            catch (hghghhhg hghghhhg8) {
                hghghhhg8.ghgggghh();
                throw hghghhhg8;
            }
            try {
                stringArray2 = new String[n];
            }
            catch (hghghhhg hghghhhg9) {
                hghghhhg9.ghgggghh();
                throw hghghhhg9;
            }
            try {
                stringArray2[0] = "65537";
            }
            catch (hghghhhg hghghhhg10) {
                hghghhhg10.ghgggghh();
                throw hghghhhg10;
            }
            try {
                stringArray = stringArray2;
                stringArray2[1] = "RSA";
            }
            catch (hghghhhg hghghhhg11) {
                hghghhhg11.ghgggghh();
                throw hghghhhg11;
            }
            try {
                string = "RSA";
            }
            catch (hghghhhg hghghhhg12) {
                hghghhhg12.ghgggghh();
                throw hghghhhg12;
            }
            try {
                stringArray[2] = string;
            }
            catch (hghghhhg hghghhhg13) {
                hghghhhg13.ghgggghh();
                throw hghghhhg13;
            }
            try {
                ghhhhh = stringArray;
                if (!bl) {
                    return;
                }
            }
            catch (hghghhhg hghghhhg14) {
                hghghhhg14.ghgggghh();
                throw hghghhhg14;
            }
        }
    }
}
```

Example output after:
```Java
public class hhgg {
    private static /* bridge */ /* synthetic */ void ggghhg() {
        boolean bl;
        try {
            bl = ggghghg;
        }
        catch (hghghhhg hghghhhg2) {
            hghghhhg2.ghgggghh();
            throw hghghhhg2;
        }
        if (!bl) {
            String string;
            String[] stringArray;
            String[] stringArray2;
            int n;
            try {
                n = 3;
            }
            catch (hghghhhg hghghhhg3) {
                hghghhhg3.ghgggghh();
                throw hghghhhg3;
            }
            try {
                stringArray2 = new String[n];
            }
            catch (hghghhhg hghghhhg4) {
                hghghhhg4.ghgggghh();
                throw hghghhhg4;
            }
            try {
                stringArray2[0] = "65537";
            }
            catch (hghghhhg hghghhhg5) {
                hghghhhg5.ghgggghh();
                throw hghghhhg5;
            }
            try {
                stringArray = stringArray2;
                stringArray2[1] = "RSA";
            }
            catch (hghghhhg hghghhhg6) {
                hghghhhg6.ghgggghh();
                throw hghghhhg6;
            }
            try {
                string = "RSA";
            }
            catch (hghghhhg hghghhhg7) {
                hghghhhg7.ghgggghh();
                throw hghghhhg7;
            }
            try {
                stringArray[2] = string;
            }
            catch (hghghhhg hghghhhg8) {
                hghghhhg8.ghgggghh();
                throw hghghhhg8;
            }
            try {
                ghhhhh = stringArray;
                if (!bl) {
                    return;
                }
            }
            catch (hghghhhg hghghhhg9) {
                hghghhhg9.ghgggghh();
                throw hghghhhg9;
            }
        }
    }
}
```

### DIFF
--- a/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op3rewriters/ExceptionRewriters.java
+++ b/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op3rewriters/ExceptionRewriters.java
@@ -348,15 +348,14 @@ public class ExceptionRewriters {
             boolean printable = false;
 
             Op03SimpleStatement last = tryStatement.getTargets().get(tryStatement.getTargets().size() - 1);
-            Iterator<Op03SimpleStatement> targets = tryStatement.getTargets().iterator();
-            Op03SimpleStatement target;
-            while (targets.hasNext()) {
-                target = targets.next();
-                if (target == last) break; // dont check the terminating catch statement
+            for (Op03SimpleStatement target : tryStatement.getTargets()) {
                 Statement statement = target.getStatement();
 
+                // dont check the terminating catch statement
+                if (target == last && statement instanceof CatchStatement) break;
+
                 if (!(
-                    statement instanceof GotoStatement
+                    (statement instanceof GotoStatement && !(statement instanceof IfStatement))
                     ||
                     statement instanceof Nop
                 )) {
--- a/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op3rewriters/ExceptionRewriters.java
+++ b/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op3rewriters/ExceptionRewriters.java
@@ -7,10 +7,7 @@ import org.benf.cfr.reader.bytecode.analysis.parse.LValue;
 import org.benf.cfr.reader.bytecode.analysis.parse.Statement;
 import org.benf.cfr.reader.bytecode.analysis.parse.expression.StackValue;
 import org.benf.cfr.reader.bytecode.analysis.parse.lvalue.StackSSALabel;
-import org.benf.cfr.reader.bytecode.analysis.parse.statement.AssignmentSimple;
-import org.benf.cfr.reader.bytecode.analysis.parse.statement.CatchStatement;
-import org.benf.cfr.reader.bytecode.analysis.parse.statement.CommentStatement;
-import org.benf.cfr.reader.bytecode.analysis.parse.statement.TryStatement;
+import org.benf.cfr.reader.bytecode.analysis.parse.statement.*;
 import org.benf.cfr.reader.bytecode.analysis.parse.utils.BlockIdentifier;
 import org.benf.cfr.reader.bytecode.analysis.parse.utils.BlockIdentifierFactory;
 import org.benf.cfr.reader.bytecode.analysis.parse.utils.BlockType;
@@ -39,7 +36,6 @@ public class ExceptionRewriters {
         }
         return statements;
     }
-
 
     private static boolean eliminateCatchTemporary(Op03SimpleStatement catchh) {
         if (catchh.getTargets().size() != 1) return false;
@@ -346,25 +342,48 @@ public class ExceptionRewriters {
         List<Op03SimpleStatement> tries = getTries(in);
         for (Op03SimpleStatement tryStatement : tries) {
             BlockIdentifier block = ((TryStatement)tryStatement.getStatement()).getBlockIdentifier();
+
+            // Check whether the contents of the try block can either throw an exception, or be dumped to output
+            // If both are false then this try catch block should be removed
+            boolean printable = false;
+
+            Op03SimpleStatement last = tryStatement.getTargets().get(tryStatement.getTargets().size() - 1);
+            Iterator<Op03SimpleStatement> targets = tryStatement.getTargets().iterator();
+            Op03SimpleStatement target;
+            while (targets.hasNext()) {
+                target = targets.next();
+                if (target == last) break; // dont check the terminating catch statement
+                Statement statement = target.getStatement();
+
+                if (!(
+                    statement instanceof GotoStatement
+                    ||
+                    statement instanceof Nop
+                )) {
+                    printable = true;
+                    break;
+                }
+            }
+
             Op03SimpleStatement tgtStm = tryStatement.getTargets().get(0);
-            if (tgtStm.getBlockIdentifiers().contains(block)) {
+            if (printable && tgtStm.getBlockIdentifiers().contains(block)) {
                 continue;
             }
+
             // Ok, it doesn't contain it.  We now have to check if that block is used anywhere :(
             if (firstByBlock == null) {
                 firstByBlock = getFirstByBlock(in);
             }
-            if (firstByBlock.containsKey(block)) {
+            if (printable && firstByBlock.containsKey(block)) {
                 continue;
             }
-            // Either remove altogether (not implemented yet), or insert a spurious statement
-            Op03SimpleStatement newStm = new Op03SimpleStatement(tryStatement.getBlockIdentifiers(),
-                    new CommentStatement("empty try"), tryStatement.getIndex().justAfter());
-            newStm.getBlockIdentifiers().add(block);
-            newStm.addSource(tryStatement);
-            newStm.addTarget(tgtStm);
-            tgtStm.replaceSource(tryStatement, newStm);
-            tryStatement.replaceTarget(tgtStm, newStm);
+
+            tgtStm.removeSource(tryStatement);
+            for (Op03SimpleStatement source : tryStatement.getSources()) {
+                source.replaceTarget(tryStatement, tgtStm);
+                tgtStm.addSource(source);
+            }
+            
             effect = true;
         }
 


cookiedragon234:
This also fixes the previous hacky implementation used to remove empty try blocks:
```Java
            // Either remove altogether (not implemented yet), or insert a spurious statement
            Op03SimpleStatement newStm = new Op03SimpleStatement(tryStatement.getBlockIdentifiers(),
                    new CommentStatement("empty try"), tryStatement.getIndex().justAfter());
            newStm.getBlockIdentifiers().add(block);
            newStm.addSource(tryStatement);
            newStm.addTarget(tgtStm);
            tgtStm.replaceSource(tryStatement, newStm);
            tryStatement.replaceTarget(tgtStm, newStm);
```

cookiedragon234:
Also I should say that this code requires the `forcetopsort` option. Maybe this could be enabled by default?

GraxCode:
@leibnitz27 That's what i meant with checking every opcode if it can possibly throw anything, when I messaged you on twitter. This way unnecessary try catch blocks can be removed. Also there are more opcodes than GOTO and NOP that can be treated as "no-side-effect".

cookiedragon234:
@GraxCode As I said in the post, I'm not sure whether to remove try blocks with printable content that cannot throw an exception. While it would produce cleaner output, it would also create inaccuracy with valid javac code. Perhaps this could be added behind an option?

GraxCode:
I would also be for an option. But if the try catch block e.g. only includes "iconst_0" or some stack op it should be removed either way. Compilers wouldn't produce such output, they would produce statements. so
```java
try {
aconst_null
} catch (...)
```
should be removed regardlessly, while
```java
try {
variable = 5 % 2;
} catch (...)
```
should only be removed with an enabled "aggressive" setting.

leibnitz27:
Can you provide some example class files?

(Ideally a PR into cfr_tests/hardcoded_class_files, as I suspect you'll be building by hand).

I would also guard these behind a dead code removal flag.  (as per:)

   --removedeadconditionals         (boolean)
   --removedeadmethods              (boolean)  default: true

leibnitz27:
[ExceptionBang2.class.txt](https://github.com/leibnitz27/cfr/files/5071101/ExceptionBang2.class.txt)

This removes legitimate and necessary exception handlers.

```
c:\code\cfr\target\classes>java org.benf.cfr.reader.Main c:\code\java\gaussian\out\production\gaussian\org\benf\gaussian\foo\ExceptionBang2.class
/*
 * Decompiled with CFR 0.151-SNAPSHOT (f27757b-dirty).
 */
package org.benf.gaussian.foo;

public class ExceptionBang2 {
    public static void main(String ... args) {
        System.setOut(null);
        try {
            System.out.println("FOOO");
        }
        catch (RuntimeException e) {
            System.err.println("SECRET MESSAGE");
        }
        System.err.println("DONE");
    }
}

c:\code\cfr\target\classes>java org.benf.cfr.reader.Main c:\code\java\gaussian\out\production\gaussian\org\benf\gaussian\foo\ExceptionBang2.class --recpass 2
/*
 * Decompiled with CFR 0.151-SNAPSHOT (f27757b-dirty).
 */
package org.benf.gaussian.foo;

public class ExceptionBang2 {
    /*
     * Enabled aggressive block sorting
     * Enabled unnecessary exception pruning
     * Enabled aggressive exception aggregation
     */
    public static void main(String ... args) {
        System.setOut(null);
        {
            System.out.println("FOOO");
        }
        System.err.println("DONE");
    }
}
```

cookiedragon234:
@leibnitz27 Could you check how you are running it? Your output says `f27757b-dirty` but it should be `a39ecf5`. 

Here's my output:
```Java
/usr/lib/jvm/java-8-openjdk/bin/java -classpath /run/media/cook/DATA/Computing/cfr/target/classes org.benf.cfr.reader.Main /home/cook/Downloads/ExceptionBang2.class
/*
 * Decompiled with CFR 0.151-SNAPSHOT (a39ecf5).
 */
package org.benf.gaussian.foo;

public class ExceptionBang2 {
    public static void main(String ... args) {
        System.setOut(null);
        try {
            System.out.println("FOOO");
        }
        catch (RuntimeException e) {
            System.err.println("SECRET MESSAGE");
        }
        System.err.println("DONE");
    }
}
```

leibnitz27:
Scroll to the right of my example - unfortunately github doesn't wrap the lines - add the parameter --recpass 2 (which just forces some recovery settings early, rather than you having to construct an example which triggers them, which can be particularly heinous).

leibnitz27:
FWIW, though, the reason I went straight to '--recpass 2' is to force topsort, as the code you've changed won't even get called unless topsort flag is set, so that's a super quick way of forcing your code to be used.

```
        if (options.getOption(OptionsImpl.FORCE_TOPSORT) == Troolean.TRUE) {
            if (options.getOption(OptionsImpl.FORCE_RETURNING_IFS) == Troolean.TRUE) {
                Op03Rewriters.replaceReturningIfs(op03SimpleParseNodes, true);
            }
            if (options.getOption(OptionsImpl.FORCE_COND_PROPAGATE) == Troolean.TRUE) {
                Op03Rewriters.propagateToReturn2(op03SimpleParseNodes);
            }
            ExceptionRewriters.handleEmptyTries(op03SimpleParseNodes); <-- your changes are in here.
```

leibnitz27:
In general, I think your logic has an issue - consider if (let's say for some reason it hadn't been removed) the first statement inside the try was a goto, that just jumped to the next statement.

Simplifying your logic a little - you're doing this

```
            for (Op03SimpleStatement target : tryStatement.getTargets()) {
                Statement statement = target.getStatement();

                // dont check the terminating catch statement
                if (target == last && statement instanceof CatchStatement) break;
                if (statement.getClass() != GotoStatement.class && !(statement instanceof Nop)) {
                    printable = true;
                    break;
                }
            }
```

which means 

```
try {
GOTO X
X: DO SOMETHING IMPORTANT WHICH THROWS
LOTS MORE STUFF
STUFF
HERE
} catch {
  HANDLE
}
```

would count as an unnecessary try block.  

This isn't really valid.    You need to verify that there isn't a single statement in the block that could throw.

cookiedragon234:
> Scroll to the right of my example - unfortunately github doesn't wrap the lines - add the parameter --recpass 2 (which just forces some recovery settings early, rather than you having to construct an example which triggers them, which can be particularly heinous).

yep, sorry reset all my dev enviroments and lost my run script which had top sorting


I see the problem, for some reason I assumed that the targets of the try block would include every contained instruction, which obviously wouldn't make sense. How would I go about getting all the contained instructions? (Sorry, still trying to get around your codebase!)

cookiedragon234:
@leibnitz27 ^

leibnitz27:
Hey - sorry, bit swamped atm.  I think this would probably work better overall in a slightly different place in the control flow, I'll have a go when I'm done with my current task and reference it here.  (essentially, at exception discovery stage, not post processing.)

Would still appreciate the example class file you used above - I have some synthetics but more examples = better ;)

cookiedragon234:
Afraid I've forgotten where I got the example from and anyway it was probably proprietary.

I created a new sample using the following radon config:
```Yml
# Input
input: "obftest.jar"

# Output
output: "obftest-OBF.jar"

# Classpath to use
libraries:
  - "/usr/lib/jvm/java-8-jdk/lib/rt.jar"

# Obfuscates the flow of the bytecode
flow_obfuscation:
  # Insert fake try-catch blocks (WARNING: will bloat JAR)
  fake_catch_blocks: true

# Level of compression to use on the output JAR
compression_level: 9

# Name generation dictionary to use (custom)
dictionary: "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"

# Initial string length for randomly generated strings
randomized_string_length: 1

# Exclusions to mask certain classes or their members from being tampered with
exclusions:
```

[ArithmeticTest.zip](https://github.com/leibnitz27/cfr/files/5171627/ArithmeticTest.zip)

cookiedragon234:
Tysm, looks great

PR #211, opened 2020-11-14
Anti-Obfuscation: Arithmetic and casting folding

Col-E:
This achieves the same effect as #210 but instead of bytecode manipulations, it instead transforms expression instances. 

***

Examples (Using `antiobf` set to `true`)
```java
// IXOR
public static int fibonacci(int n) {
    if (n <= (0x67DA ^ 0xFFFFE265 ^ 0x6184 ^ 0xFFFFE43A)) {
        return n;
    }
    return fibonacci(n - (0xBBB ^ 0xFFFFE60B ^ 0x65F2 ^ 0xFFFF8843)) + fibonacci(n - (0x7AD6 ^ 0xFFFFA6F2 ^ 0x66B2 ^ 0xFFFFBA94));
}
// ISUB + IADD
public static int fibonacci(int n) {
    if (n <= 36 - 40 + 26 - 15 - 6) {
        return n;
    }
    return fibonacci(n - (208 - 355 + 353 - 245 + 40)) + fibonacci(n - (223 - 412 + 193 - 2));
}
// I2L + LXOR + L2I
public static int fibonacci(int n) {
    if (n <= (int)((long)2108268340 ^ (long)2108268341)) {
        return n;
    }
    return fibonacci(n - (int)((long)864501408 ^ (long)864501409)) + fibonacci(n - (int)((long)1741113313 ^ (long)1741113315));
}
```
Cleaned output
```java
public static int fibonacci(int n) {
    if (n <= 1) {
        return n;
    }
     return fibonacci(n - 1) + fibonacci(n - 2);
}
```

Samples: 

- https://github.com/GenericException/SkidSuite/blob/master/obf/Scuti/Num2/output.jar
- https://github.com/GenericException/SkidSuite/blob/master/obf/Radon/samples/1.0.3/Num1/output.jar
- https://github.com/GenericException/SkidSuite/blob/master/obf/Radon/samples/1.0.3/Num3/output.jar


leibnitz27:
copying comment from other PR:

An expression transformer can be applied either at 3rd (graph) stage
or 4th (fully structured) stage.

For 3rd, look at NOPSearchingExpressionRewriter
For 4th, look at HexLiteralTidier

Col-E:
These get applied after `HexLiteralTidier`

leibnitz27:
Thanks -  Couple of things - 

1)
You don't need a separate rewriter for casts - just allow the one rewriter to hit everything.

```
	@Override
	public Expression rewriteExpression(Expression expression, SSAIdentifiers ssaIdentifiers, StatementContainer statementContainer, ExpressionRewriterFlags flags) {
		expression.applyExpressionRewriter(this, ssaIdentifiers, statementContainer, flags);
                // TODO: Probably only worth doing if type of expression is foldable, i.e. if it's a RawJavaType.
		Expression newMe = expression.getComputedLiteral(STATIC_READONLY_MAP);
		return newMe == null ? expression : newMe;
	}
```

2)
The map is used to allow known local variables to be used in folding elsewhere (getComputedLiteral is mainly used in determining if branch tests are tautologies) - since you're never passing anything in, you can avoid the alloc on every call.

3)
Also - can you please rename ConstInlinerRewriter -> ConstantFoldingRewriter? 

4)
I'm not convinced about assuming all arithmetic is equal - I think it's liable to be better to implement arithmetic over all types explicitly, I'm afraid.  (This is where c++ templates would be handy, natch.)

Consider:
[FoldTest.txt](https://github.com/leibnitz27/cfr/files/5542181/FoldTest.txt)

this is definitely wrong, but there are more subtle ones around shorts etc.

Cheers!

Col-E:
1. Wasn't sure what would be preferred, assumed separation of duties

2. Ah, is there a way to pass in local variables that are used similar to temporary constants? I'd like to see if being hyper aggressive with folding could fold synthetic variables. If not, then yeah making it a constant. 

3. 👍 

4. I'll make some more test cases. Any other edge cases to consider?

leibnitz27:
> 1. Wasn't sure what would be preferred, assumed separation of duties

Yeah, fair enough, but I think both of these really are the same thing, just viewed from a slightly different angle - feels like it's not worth splitting out.

> 2. Ah, is there a way to pass in local variables that are used similar to temporary constants? I'd like to see if being hyper aggressive with folding could fold synthetic variables. If not, then yeah making it a constant.

TBH I'd really not consider that now - you need to guarantee state of variables on entry, and it's a giant pain.  Certainly wouldn't consider for this PR ;)  (Also, it's much easier to do it at the Op03 stage, where you have a graph rather than structured code).

> 4. Agreed. I'll make some more test cases. Any other edge cases to consider?

Can't think of anything specific, but would definitely worry about edge of representational range cases, or cases where float/double rounding error differs in interesting ways.

for float/double, for example, you might look at (the unfolded versions of:)

```
        System.out.println(1.23f + 3.45f);
        System.out.println(1.23 + 3.45);
```

A ridiculous-but-technically-possible edge case is where you fold doubles under non strictfp conditions, and the method is marked as strictfp.   I've never (not that I've particularly tried too hard) found an example where this makes a difference, (JVM spec doesn't allow something like FMA, which is the usual candidate for intermediate representational differences).  (if you have such an example, I'd love to see it).

Col-E:
> I'm not convinced about assuming all arithmetic is equal

Yeah, went ahead and did that. Regarding your folding example it will now print `4.6800003f`

> A ridiculous-but-technically-possible edge case is where you fold doubles under non strictfp conditions, and the method is marked as `strictfp`

In the literal computation for arithmetic, how difficult would it be to pass to an expression being evaluated that it is inside a `strictfp` method? You could then flip the computation between a `strictfp` copy of the `float/double` methods. Duplicate code sucks but its the only idea that popped in my head.

> Also, it's much easier to do it at the Op03 stage, where you have a graph rather than structured code

Something about stages I noticed, the expression folding breaks array resugaring. For example:

```java
// was
return Objects.hash(this.x, this.y);
// becomes
Object[] objectArray = new Object[2];
objectArray[0] = this.x;
objectArray[1] = this.y;
return Objects.hash(objectArray);
```

To fix this, would the transformer have to be updated to run in a prior stage? If so, got any references/pointers I could look at?

Lanchon:
IMHO there is no reason not to always constant-fold with strictfp.

Col-E:
> IMHO there is no reason not to always constant-fold with strictfp.

So would it be more sensible to be consistent than conditionally accurate?

leibnitz27:
Is there any particular reason you're checking specific operations?   You could compress all of

```
		// Simplify arithmetic
		if (expression instanceof ArithmeticOperation || expression instanceof ArithmeticMonOperation) {
			Expression computed = expression.getComputedLiteral(getDisplayMap());
			if (computed != null) {
				expression = computed;
			}
		}
		// Simplify casts for cases like "(int) 0" into just "0"
		else if (expression instanceof CastExpression) {
			Expression computed = expression.getComputedLiteral(getDisplayMap());
			if (computed != null) {
				expression = computed;
			}
		}
```
into 
```
			Expression computed = expression.getComputedLiteral(getDisplayMap());
			if (computed != null) {
				expression = computed;
			}
```

If the expression doesn't support folding it will return quickly, and if it does, free work.

Are there any cases this messes up?

src/org/benf/cfr/reader/bytecode/analysis/parse/rewriters/ConstantFoldingRewriter.java
```diff
@@ -0,0 +1,54 @@
...
+	private static final Map<LValue, Literal> DISPLAY_MAP = MapFactory.<LValue, Literal>newMap();
+
+	@Override
+	public Expression rewriteExpression(Expression expression, SSAIdentifiers ssaIdentifiers, StatementContainer statementContainer, ExpressionRewriterFlags flags) {
+		expression.applyExpressionRewriter(this, ssaIdentifiers, statementContainer, flags);
+		// Skip if expression type is non-primitive
+		if (expression.getInferredJavaType().getRawType().ordinal() > RawJavaType.DOUBLE.ordinal())
+			return expression;
+		// Simplify arithmetic
+		if (expression instanceof ArithmeticOperation || expression instanceof ArithmeticMonOperation) {
```
leibnitz27, line 27: Is there any particular reason for the explicit tests?  

```
			Expression computed = expression.getComputedLiteral(getDisplayMap());
			if (computed != null) {
				expression = computed;
			}
```

would work just as well, and be more general - if something doesn't support it, it will return quickly...

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/CastExpression.java
```diff
@@ -57,6 +63,34 @@ public Expression deepClone(CloneHelper cloneHelper) {
         return new CastExpression(getLoc(), getInferredJavaType(), cloneHelper.replaceOrClone(child), forced);
     }
 
+    @Override
+    public Literal getComputedLiteral(Map<LValue, Literal> display) {
+        if (getChild() instanceof Literal) {
+            Literal childLiteral = (Literal) getChild();
+            CastAction action = childLiteral.getValue().getInferredJavaType().chain(getInferredJavaType());
```
leibnitz27, line 70: You don't need to chain here - type chaining (that code is very confusing) is used to generate type dependencies.

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/CastExpression.java
```diff
@@ -57,6 +63,34 @@ public Expression deepClone(CloneHelper cloneHelper) {
         return new CastExpression(getLoc(), getInferredJavaType(), cloneHelper.replaceOrClone(child), forced);
     }
 
+    @Override
+    public Literal getComputedLiteral(Map<LValue, Literal> display) {
+        if (getChild() instanceof Literal) {
```
leibnitz27, line 68: You *can* check for a literal here, but it's probably more general just to getComputedLiteral of child, and see what that gets you.

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/CastExpression.java
```diff
@@ -57,6 +63,34 @@ public Expression deepClone(CloneHelper cloneHelper) {
...
+            Literal childLiteral = (Literal) getChild();
+            CastAction action = childLiteral.getValue().getInferredJavaType().chain(getInferredJavaType());
+            if (action == CastAction.None) {
+                return childLiteral;
+            }
+            Number value = (Number) childLiteral.getValue().getValue();
+            switch (getInferredJavaType().getRawType()) {
```
leibnitz27, line 75: (plus other places)

Rather than switching on getRawType, can you instanceof getJavaTypeInstance(), cast to RawJavaType (and fail-return-null if not), and switch on that?

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/CastExpression.java
```diff
@@ -57,6 +63,34 @@ public Expression deepClone(CloneHelper cloneHelper) {
...
+                case BOOLEAN:
+                case CHAR:
+                case SHORT:
+                case INT:
+                case BYTE:
+                    return new Literal(TypedLiteral.getInt(value.intValue()));
```
leibnitz27, line 81: This will have the wrong type inside it - perhaps add a getInt(intVal, RawJavaType) to TypedLiteral

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/ArithmeticOperation.java
```diff
@@ -90,6 +92,174 @@ public Dumper dumpInner(Dumper d) {
...
+                res = lv + rv;
+                break;
+            case MINUS:
+                res = lv - rv;
+                break;
+            case MULTIPLY:
+                res = lv * rv;
+                break;
+            case DIVIDE:
+                res = lv / rv;
```
leibnitz27, line 138: This will make the decompiler fail with a / 0 error

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/ArithmeticOperation.java
```diff
@@ -90,6 +92,174 @@ public Dumper dumpInner(Dumper d) {
...
+                break;
+            case REM:
+                res = lv % rv;
```
leibnitz27, line 141: / 0 exception

Lanchon:
> > IMHO there is no reason not to always constant-fold with strictfp.
> 
> So would it be more sensible to be consistent than conditionally accurate?


absolutely. there is no reason to inject math inaccuracies of the runtime running CFR into the resulting source code (that might later be ran on a different paltform). it is customary to do constant folding as precisely as possible.

Lanchon:
not to meantion that CFR runs would become not repeatable (bad for testing).

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/ArithmeticOperation.java
```diff
@@ -90,6 +92,174 @@ public Dumper dumpInner(Dumper d) {
...
+        Number lv = ((Number) l.value.getValue());
+        Number rv = ((Number) r.value.getValue());
+        RawJavaType widestType = lType.ordinal() > rType.ordinal() ? lType : rType;
+        switch (widestType) {
+            case BOOLEAN:
+            case BYTE:
+            case CHAR:
+            case SHORT:
+            case INT:
+                return getComputedInt(rv.intValue(), lv.intValue());
```
leibnitz27, line 114: I just don't think treating all things as integers is safe - 

[CastTest2.txt](https://github.com/leibnitz27/cfr/files/5549380/CastTest2.txt)

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/ArithmeticMonOperation.java
```diff
@@ -58,6 +62,83 @@ public Dumper dumpInner(Dumper d) {
         return d;
     }
 
+    @Override
+    public Literal getComputedLiteral(Map<LValue, Literal> display) {
+        Literal l = lhs.getComputedLiteral(display);
+        if (l == null || !(l.getValue().getValue() instanceof Number))
```
leibnitz27, line 68: as silly as it seems, I'd be tempted to avoid the Number path, and use the type that the Literal tells you it is.  I suspect it may be a bit more boiler plate, but it's probably less liable to lead to odd typing edge cases.

leibnitz27:
@Col-E  - pls don't take the volume of comments personally, it's just how I roll ;)

If you like, I can tinker with your branch to show you what I mean... (or will leave with you if you prefer)

leibnitz27:
I know I'm the one who uncorked the bottle of the strictfp genie, but I'd really like to know if there are examples (source for earlier image?) where it actually makes a difference?    TBH - I would never enable this sort of optional behaviour by default.... so I'm not HUGELY worried, just..... a bit obsessive..... ;)

Col-E:
> Is there any particular reason you're checking specific operations? You could compress all of...

It expresses the train of thought more accurately since these few items are the only items within scope.  But its correct that this is not necessary.

> TBH - I would never enable this sort of optional behaviour by default.... so I'm not HUGELY worried, just..... a bit obsessive..... ;)

I'm the type who would and would like to implement it in my tools if merged :)
And I totally understand that obsession.

> pls don't take the volume of comments personally, it's just how I roll ;)

👍 

> If you like, I can tinker with your branch to show you what I mean... (or will leave with you if you prefer)

I'd love some examples if you have anything handy. 

>  it is customary to do constant folding as precisely as possible.

Understandable, I think you meant consistent though. AFAIK StrictFP loses some precision vs system-dependant fp operations.

src/org/benf/cfr/reader/bytecode/analysis/parse/rewriters/ConstantFoldingRewriter.java
Col-E: No real reason. I just wanted to avoid potentially unknown behavior outside of the scope of this PR. I can skip the checks.

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/CastExpression.java
Col-E: Ah, I'm using this because I assumed this would allow to check if a cast is redundant easily _(by checking against none)_

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/CastExpression.java
Col-E: Fair, I assumed the instance check would be more performant than potentially a failed computation. I'd have to look at what implements this to see possible effects on that. This was just a safe assumption IMO, can be changed though.

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/ArithmeticOperation.java
Col-E: How should we handle these sorts of cases gracefully? Just ignore computation?

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/ArithmeticMonOperation.java
Col-E: I used this as a quick way to assert that a value was there of a supported type since `instanceof` also null checks. It means I can safely cast later too, so I'd assume I've have to make this test somewhere regardless.

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/ArithmeticOperation.java
Col-E: I believe this is correct as it falls under primitive widening. The issue with the `CastTest2` seems to be improperly handling the value in the `CastExpression` computation.

Col-E: I've pushed the fix for this in `CastExpression` where `CastTest2.txt` now prints `-2` as the computed byte value.

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/ArithmeticMonOperation.java
```diff
@@ -58,6 +62,83 @@ public Dumper dumpInner(Dumper d) {
...
+            return null;
+        Number lv = ((Number) l.value.getValue());
+        switch (l.getInferredJavaType().getRawType()) {
+            case BOOLEAN:
+            case BYTE:
+            case CHAR:
+            case SHORT:
+            case INT:
+                return getComputedInt(lv.intValue());
```
Col-E, line 77: May want to create some test cases to validate if mon operations on narrower types can get messed up due to treating as integers. The solution for the bi-operations was to simply cast afterwards. If the same applies here then the solution is simple.

leibnitz27:
I've pushed https://github.com/leibnitz27/cfr/tree/Col-E-inlining which contains most of the tweaks I mention above - might be useful.

[CastTest3.txt](https://github.com/leibnitz27/cfr/files/5555543/CastTest3.txt)

This class file also contains a few tests

leibnitz27:
Ah, didn't see your latest push when I wrote that.  ;)

Col-E:
Oh but moving the literal computation is a [nice idea](https://i.ytimg.com/vi/t0sTNLdNhuE/hqdefault.jpg).

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/ArithmeticMonOperation.java
```diff
@@ -58,6 +62,83 @@ public Dumper dumpInner(Dumper d) {
...
+            case LONG:
+                return getComputedLong(lv.longValue());
```
Col-E, line 79: Also making a note of here and in the other class to move the computation to a new class to hide some of the repetitive code in these expression classes.

leibnitz27:
Let me know when you're done (sounds like you want to do a bit from last comment)

Col-E:
I'm away for a bit _(3-4 hours or so)_ so you can review the current changes until I can get back. I marked items as resolved as I went but you can double check.

I see in the example push you separated the arith for narrow primitives, and I've not yet done that.

leibnitz27:
no hurry!

Col-E:
```java
// before
CastTest3.foo((byte)(127 + 127));
CastTest3.foo((float)((double)1.23f + (double)3.45f));
CastTest3.foo((double)1.23f + (double)3.45f);
CastTest3.foo(1.23f + 3.45f);
CastTest3.foo((byte)(127 / 0));
// after
CastTest3.foo(-2);
CastTest3.foo(4.6800003f);
CastTest3.foo(4.680000066757202);
CastTest3.foo(4.6800003f);
CastTest3.foo((byte)(127 / 0));
```
Tested by pasting the expressions from `before` into IJ's evaluator and got the exact matches of the `after` so these ought to be correct.

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/ArithmeticMonOperation.java
Col-E: Doesn't cause problems. Compared by checking against:
```java
// foo = println
foo((~(short)333));
foo((~(short)222));
foo((~(short)111));
foo((~(short)-111));
foo((~(short)-333));
System.out.println();
foo((~(byte)333));
foo((~(byte)222));
foo((~(byte)111));
foo((~(byte)-111));
foo((~(byte)-333));
System.out.println();
```
The corresponding values printed when running:
```
-334
-223
-112
110
332

-78
33
-112
110
76
```
The decompiled output:
```java
Mon.foo(-334);
Mon.foo(-223);
Mon.foo(-112);
Mon.foo(110);
Mon.foo(332);
System.out.println();
Mon.foo(-78);
Mon.foo(33);
Mon.foo(-112);
Mon.foo(110);
Mon.foo(76);
```

src/org/benf/cfr/reader/bytecode/analysis/parse/expression/ArithmeticOperation.java
Col-E: Marking this as resolved with the inclusion of some of the recent changes _(and paste from your example)_

src/org/benf/cfr/reader/bytecode/analysis/parse/rewriters/ConstantFoldingRewriter.java
```diff
@@ -0,0 +1,50 @@
...
+
+public class ConstantFoldingRewriter extends AbstractExpressionRewriter {
+	public static final ConstantFoldingRewriter INSTANCE = new ConstantFoldingRewriter();
+	private static final Map<LValue, Literal> DISPLAY_MAP = MapFactory.<LValue, Literal>newMap();
+
+	@Override
+	public Expression rewriteExpression(Expression expression, SSAIdentifiers ssaIdentifiers, StatementContainer statementContainer, ExpressionRewriterFlags flags) {
+		expression.applyExpressionRewriter(this, ssaIdentifiers, statementContainer, flags);
+		// Skip if expression type is non-primitive
+		JavaTypeInstance type = expression.getInferredJavaType().getJavaTypeInstance();
```
leibnitz27, line 25: Doesn't actually skip if not RawJavaType, will add ;)

leibnitz27:
Cool!

Col-E:
🥳

PR #215, opened 2020-11-29
Make project can compile on jdk8

boroborome:


pom.xml
```diff
@@ -174,7 +174,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-        <javaVersion>1.6</javaVersion>
+        <javaVersion>1.8</javaVersion>
```
leibnitz27, line 177: Sorry, absolutely not - this project is 1.6 very deliberately.

leibnitz27: You can always override the compiler if you need to.

leibnitz27:
No objection to making it more J8 friendly - https://github.com/leibnitz27/cfr/commit/0931cbfcaf266ca4891c63ed1fd589a7f7845756 removes that compile error (in a strongly typed way - pls don't change signatures to be weakly typed).

But this project is staying 1.6 for entirely illogical reasons that it's more of a challenge that way. \o/

leibnitz27:
http://www.benf.org/other/cfr/faq.html#java6 btw ;)

PR #229, opened 2021-02-22
add external classFileSource to the CFR driver

amirrapson:
Added an external file source to the interface, to allow CFR to find dependent classes that are not part of the current classpath or in an added jar.

### DIFF
--- a/src/org/benf/cfr/reader/CfrDriverImpl.java
+++ b/src/org/benf/cfr/reader/CfrDriverImpl.java
@@ -45,6 +45,11 @@ public class CfrDriverImpl implements CfrDriver {
 
     @Override
     public void analyse(List<String> toAnalyse) {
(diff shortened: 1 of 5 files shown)


Col-E:
This isn't really necessary, you can use `ClassFileSource#getClassFileContent`

Example impl: [ClassSource.java](https://github.com/Col-E/Recaf/blob/master/src/main/java/me/coley/recaf/decompile/cfr/ClassSource.java)

```java
ClassSource source = new ClassSource(getController());
SinkFactoryImpl sink = new SinkFactoryImpl();
CfrDriver driver = new CfrDriver.Builder()
		.withClassFileSource(source)
		.withOutputSink(sink)
		.withOptions(getOptions())
		.build();
driver.analyse(Collections.singletonList(name));
```

PR #244, opened 2021-04-19
Fix javadocs on annotation classes

YanisBft:
Javadocs were not showing in annotation classes, this PR fixes it


leibnitz27:
Hey - sorry for massively late reply - the dumpComments method doesn't do javadocs, it dumps any observations about the annotation.  That being said, there's definitely no reason not to do that here ;)

PR #260, opened 2021-07-16
add --outputencoding flag,supports saving decompiled files with specified encoding

passedbylove:
add --outputencoding flag,supports saving decompiled files with specified encoding for non-english speaking country.

eg. --outputencoding UTF-8

### DIFF
--- a/src/org/benf/cfr/reader/util/getopt/OptionsImpl.java
+++ b/src/org/benf/cfr/reader/util/getopt/OptionsImpl.java
@@ -452,6 +452,9 @@ public class OptionsImpl implements Options {
     public static final PermittedOptionProvider.Argument<String> OUTPUT_PATH = register(new PermittedOptionProvider.Argument<String>(
             "outputpath", defaultNullStringDecoder,
             "Decompile to files in [directory]"));
+    public static final PermittedOptionProvider.Argument<String> OUTPUT_ENCODING = register(new PermittedOptionProvider.Argument<String>(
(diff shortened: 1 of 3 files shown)


passedbylove:
modern operation system were using Unicode Character Table,in my opion,saving decompiled class with utf-8 character will be the good choise.

x4e:
Would it be better to store the charset as an instance of java.nio.charset.Charset rather than a String?

passedbylove:
hi guy,acrodding to your suggesting ,i've found that signature
`BufferedWriter(OutputStreamWriter, java.nio.Charset)) ` will not need try-catch block, besides this, are there any deiffrent between
`BufferedWriter(OutputStreamWriter, java.nio.Charset))` and `BufferedWriter(OutputStreamWriter, String))` ?

passedbylove:
> Would it be better to store the charset as an instance of java.nio.charset.Charset rather than a String?

hi guy,acrodding to your suggesting ,i've found that signature
`BufferedWriter(OutputStreamWriter, java.nio.Charset)) ` will not need try-catch block, besides this, are there any deiffrent between
`BufferedWriter(OutputStreamWriter, java.nio.Charset))` and `BufferedWriter(OutputStreamWriter, String))` ?

leibnitz27:
Seems reasonable, thanks; as x4e says, minor optimisation as Charset, but in the grand scheme of things it's hardly the biggest efficiency offender, can be tweaked ;)

Issue #172, opened 2020-05-20
Add `version` command line option


Marcono1234:
# Problem solved by this feature
Commit 40cfc84 changed it so the version information does not include the commit hash anymore if the version indicates that the build is a release.
This can make handling GitHub bug reports more difficult if people checked out the release and then modified the source:
- If they do not commit their changes, the version information will not indicate that their working tree is dirty
- If they do commit their changes, but do not change the CFR version (so it remains on the release version), the version information will not indicate this

I was thinking about adding a `version` command line option which always includes the commit hash. And then adjust the `bug_report.md` GitHub issue template to say that this option should be used to get the version information.

The problem is that this `version` option is not really an option, the same way `--help` is not an option either. I would rather consider them "commands" because they also work without argument (for `version` it only works without argument) and combining them with other options makes no sense, though the similar syntax suggests it would:
<pre>
java -jar cfr.jar --labelledblocks true <b>--help test</b> --override false my.jar
</pre>

# Feature description
Do not implement `help` and `version` as regular options, but as commands. To prevent ambiguities a `decompile` command should be added which acts the same way the current command line parsing does.
`help` without arguments should list all commands (and for `decompile` all options). And `help` with argument should only use the argument to look up `decompile` options, but not commands.

Examples:

- `java -jar cfr.jar help`
- `java -jar cfr.jar help labelledblocks`
- `java -jar cfr.jar version`
- `java -jar cfr.jar decompile --override false my.jar`

# Considerations
- It breaks command line backward compatibility since now you have to prefix the command with `decompile`
- You would have to type an additional word `decompile` when decompiling
- The connection between the `help` and `decompile` command might not be obvious enough

# Alternatives
Introduce special casing and hacky workarounds?

Lanchon:
sorry but big thumbs down to this change. never break unnecessarily. everybody will hate CFR for it. and '--version' is the standard tried and true way of doing this.

i'm all for the added version info though, just not the API change.

leibnitz27:
Hey - 

so - I **_*really*_** don't like the addition of `decompile`.   (no surprises there ;) ) - as such I really don't want to go down that route - would significantly prefer to stick with arguments, even if semantically it's a little fluffy,

I'm not *ENORMOUSLY* averse to reverting https://github.com/leibnitz27/cfr/commit/40cfc84940c6c64105071faa50490b6c8b9f35c8 for the next release.

[...]

PR #264, opened 2021-08-28
Minor changes: add the --version option and slightly improve the help hint

ildar-shaimordanov:
I am not sure that my suggestions are fully covered by the project contributing guideline. Nevertheless I'd like to post them here. In fact, there are two small commits:

* add --version option
  Many tools (and JVM itself) have the option for printing its version. I hope it can be useful for CFR as well.
* unify help hint outputs
  While working on the first commit I found that two hints look almost identical with some differences. So I combined them into a single method.

### DIFF
--- a/src/org/benf/cfr/reader/Main.java
+++ b/src/org/benf/cfr/reader/Main.java
@@ -45,6 +45,11 @@ public class Main {
             return;
         }
 
+        if (options.optionIsSet(OptionsImpl.VERSION)) {
+            getOptParser.showVersion();
+            return;
+        }
+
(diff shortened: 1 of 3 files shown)


Marcono1234:
Related to / resolves  #172

ildar-shaimordanov:
@Marcono1234, not exactly the same. One of my suggestions consists of a new command line option to show the actual version as follows:
```
$ java -jar target/cfr-0.152-SNAPSHOT.jar --version
CFR 0.152-SNAPSHOT (2e4addd)

```

leibnitz27:
Sorry for the delay - been workswamped \o/

PR #267, opened 2021-09-16
Fix #270:MatchIterator's  ArrayIndexOutOfBoundsException,NullPointerException

### LINKED ISSUE #270: MatchIterator throws "ArrayIndexOutOfBoundsException  and the code is not robust
# CFR Version
0.152-SNAPSHOT (44bddad)
# Compiler
OpenJDK jdk-16.0.2
# Description
1. **_ArrayIndexOutOfBoundsException for the toString () method of Matchiterator_**:
IDEA in Debug mode,Variable'view will call toString(),but this time idx==-1,
so this method"toString()" will  throw ArrayIndexOutOfBoundsException.When the advance() method 
has not been called,the idx has always been -1.
2. **_NullPointerException for the Matchiterator's hasnext(), isfinished() method_**:
The data.size() method will throw NullPointerException,Although in the current code, 
this problem will not be raised,because the constructor parameter(List data) of MatchIterator is 'ListFactory.newList()',
So so far, The data of MatchIterator won't be null.But other places use this class Matchiterator,Inevitably, 
it will not cause NullPointerException.In terms of code robustness,data!=null is necessary

# Example
``` java
//src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/SwitchStringRewriter.java
//Line:#137
private void rewriteComplex(List<StructuredStatement> structuredStatements) {
        // Rather than have a non-greedy kleene star at the start, we cheat and scan for valid start points.
        // switch OB (case OB (if-testalternativevalid OB assign break CB)* if-notvalid break assign break CB)+ CB
        MatchIterator<StructuredStatement> mi = new MatchIterator<StructuredStatement>(structuredStatements);
```
error

``` java
//src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/matchutil/MatchIterator.java
public class MatchIterator<T> {
    private final List<T> data;
    private int idx;

    public MatchIterator(List<T> data) {
        this.data = data;
        this.idx = -1;
    }
	
	public boolean hasNext() {
        return idx < data.size() - 1;
		}
    private boolean isFinished() {
        return idx >= data.size();
	}

    private MatchIterator(List<T> data, int idx) {
        this.data = data;
        this.idx = idx;
    }
    public String toString() {
        if (isFinished()) return "Finished";
        T t = data.get(idx);
        return t == null ? "null" : t.toString();
	}
}
```

chenggwang:
1.T t = data.get(idx);When the advance() method has not been called,
            idx=-1,**ArrayIndexOutOfBoundsException.**   e.g:IDEA in Debug mode,
            Variable'view will call toString().so fix this bug ,and Enriched
            the method
2.when data=null. Constructor‘s parameter(List<T> data) is 'ListFactory.newList()',
            At present, the use of it in the code will not cause
            the **_NullPointerException_**.
            But it's used elsewhere,that parameter is null,NullPointerException Will appear.
            Check null in these method {hasNext(),isFinished(),toString()}
Close #270

### DIFF
--- a/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/matchutil/MatchIterator.java
+++ b/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/matchutil/MatchIterator.java
@@ -92,9 +92,11 @@ public class MatchIterator<T> {
             }
 
         }else{
+            //data = null
             return null;
         }
-        return null;
+        //when data.size()==0
+        return "data:[]";
     }
 
     public void rewind() {
--- a/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/matchutil/MatchIterator.java
+++ b/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/matchutil/MatchIterator.java
@@ -36,11 +36,11 @@ public class MatchIterator<T> {
     }
 
     public boolean hasNext() {
-        return idx < data.size() - 1;
+        return data!=null && idx < data.size() - 1;
     }
 
     private boolean isFinished() {
-        return idx >= data.size();
+        return data!=null && idx >= data.size();
     }
 
     public boolean advance() {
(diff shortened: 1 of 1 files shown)


leibnitz27:
Hey - thanks for raising this.  

Dumping the whole thing would be very unwieldy most of the time for me - these match iterators are used against flattened trees which can be enormous.

Re the null checks - I'd rather not hide bad parameters.  Yes, this codebase is $%£$$%£ poor when it comes to null checking; and I've (blame my j6 weirdness) not done nullability decoration.  

Will merge and tweak as above.

PR #269, opened 2021-09-16
Output record type parameters when dumping record classes

### LINKED ISSUE #268: Decompiled records ignore type parameters
# CFR version
CFR 0.151, from https://www.benf.org/other/cfr/cfr-0.151.jar

# Compiler
javac 17, from oracle's non-open source licensed jdk 17

# Description
I expect a record with type parameters to be decompiled with the type parameters, yet it isn't.

# Example
`Bod.java`
```java
import java.util.*;

public record Bod<T extends Comparable<T>>(T contained) {
	public record Son<T extends Comparable<T>>(Bod<T> stuff) {}
}
```
The `javap` output shows the signature is preserved (in constant pool as `<T::Ljava/lang/Comparable<TT;>;>Ljava/lang/Record;`), yet for the `Bod` class (also `Son` nested class), the output is
```
/*
 * Decompiled with CFR 0.151.
 */
public record Bod(T contained) {

    public record Son(Bod<T> stuff) {
    }
}
```

liach:
This method seems to dump type parameters and type annotations in addition. The side effects should not be too bad, I guess...

Closes #268

### DIFF
--- a/src/org/benf/cfr/reader/entities/classfilehelpers/ClassFileDumperRecord.java
+++ b/src/org/benf/cfr/reader/entities/classfilehelpers/ClassFileDumperRecord.java
@@ -25,7 +25,9 @@ public class ClassFileDumperRecord extends AbstractClassFileDumper {
     private void dumpHeader(ClassFile c, InnerClassDumpType innerClassDumpType, Dumper d) {
         d.keyword(getAccessFlagsString(c.getAccessFlags(), dumpableAccessFlagsClass));
 
-        d.keyword("record ").dump(c.getClassType()).print("(");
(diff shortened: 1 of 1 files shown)


liach:
Hmm, when I decompile, the outer record is treated as a regular class. I need to check if this is my patch or 0.152-snapshot bug.

liach:
Might need some tweaks, as we should dump with marking this as record class definition than a regular reference.

leibnitz27:
Hey - sorry been quite busy last [forever] \o/

Looks like the reversion to regular class is due to more paranoid type checking - think I need to unwind that slightly

PR #286, opened 2022-01-24
Dump module import statements and annotations

Marcono1234:
Note: This also includes implementing `toString()` for some classes; if you want I can omit that change.

Before:
```java
module test {
    provides List with MyList;

}
```

After:
```
/*
 * Decompiled with CFR 0.153-SNAPSHOT (02dac77).
 * 
 * Could not load the following classes:
 *  test.MyList
 */
import java.util.List;
import test.MyList;

@Deprecated
module test {
    provides List with MyList;
}
```


src/org/benf/cfr/reader/entities/classfilehelpers/ClassFileDumperModule.java
```diff
@@ -108,7 +123,6 @@ private void dumpProvides(ConstantPool cp, Dumper d, List<AttributeModule.Provid
             }
             d.endCodeln();
         }
-        d.newln();
     }
```
Marcono1234, line 127: test

leibnitz27:
Nice - thanks.

PR #287, opened 2022-01-26
Module dumping improvements

Marcono1234:
- Properly dumps `requires static` (instead of including `/* static phase */`)
- Dumps `uses` directives (they were missing before)
- Dumps information about module versions as comments, if present

Example:
```java
module test {
    requires static java.sql;
    
    uses java.util.List;
}
```
Compile with `javac --module-version=99 .\module-info.java`.

Output:
```java
/*
 * Decompiled with CFR 0.153-SNAPSHOT (a260598).
 */
import java.util.List;

open module test {
    // version: 99

    requires static java.sql /* version: 17 */;

    uses List;

}
```

### DIFF
--- a/src/org/benf/cfr/reader/entities/attributes/AttributeModule.java
+++ b/src/org/benf/cfr/reader/entities/attributes/AttributeModule.java
@@ -178,6 +178,9 @@ public class AttributeModule extends Attribute {
             return offset;
         }
 
+        public int getIndex() {
+            return index;
+        }
     }
 
     public static class Provide {
--- a/src/org/benf/cfr/reader/entities/attributes/AttributeModule.java
+++ b/src/org/benf/cfr/reader/entities/attributes/AttributeModule.java
@@ -2,6 +2,7 @@ package org.benf.cfr.reader.entities.attributes;
 
 import org.benf.cfr.reader.entities.constantpool.ConstantPool;
 import org.benf.cfr.reader.entities.constantpool.ConstantPoolEntryModuleInfo;
+import org.benf.cfr.reader.entities.constantpool.ConstantPoolEntryUTF8;
 import org.benf.cfr.reader.util.bytestream.ByteData;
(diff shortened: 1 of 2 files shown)


leibnitz27:
I'm trying to think of a clean way of testing multiple module specs in
current cfrtest project - any improvement on adding a bunch of top level
source dirs and then including them in every j14+ build?

leibnitz27:
\o/ thanks!

Marcono1234:
> I'm trying to think of a clean way of testing multiple module specs in
current cfrtest project - any improvement on adding a bunch of top level
source dirs and then including them in every j14+ build?

Not really. Is that even possible or does Maven then complain about too many module declarations?

One really hacky solution might be to specify multiple Maven Compiler Plugin executions each with different inclusions for `module-info.java` files, and then passing the `javac -d` option to specify a custom output directory, e.g.:
```xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-compiler-plugin</artifactId>
  <version>3.9.0</version>
  <configuration>
    <includes>
      <!-- Specify no include for default compile -->
      <include></include>
    </includes>
  </configuration>
  <executions>
    <execution>
      <id>compile1</id>
      <goals>
        <goal>compile</goal>
      </goals>
      <configuration>
        <includes>
          <include>first/module-info.java</include>
        </includes>
        <compilerArgs>
          <arg>-d</arg>
          <arg>custom-out</arg>
        </compilerArgs>
      </configuration>
    </execution>
    <execution>
      <id>compile2</id>
      <goals>
        <goal>compile</goal>
      </goals>
      <configuration>
        <includes>
          <include>second/module-info.java</include>
        </includes>
        <compilerArgs>
          <arg>-d</arg>
          <arg>custom-out2</arg>
        </compilerArgs>
      </configuration>
    </execution>
  </executions>
</plugin>
```
But I wouldn't be surprised if that breaks in the future.

Issue #252, opened 2021-05-14
Enums are not desugared since Java 15


Marcono1234:
# CFR version
bc8d9a2

# Compiler
javac 16

# Description
Sine Java 15 enums have a method `$values()` which creates the array containing the enum constants, see [JDK-8241798](https://bugs.openjdk.java.net/browse/JDK-8241798).

This causes CFR's `EnumClassRewriter` to consider the enum to be malformed and prevents it from desugaring it.

# Example
Any enum compiled with Java >= 15 should allow reproducing this, e.g.:
```java
enum EnumTest {
    A,
    B;
}
```

PR #290, opened 2022-02-12
Fix empty Java 15+ enum being dumped as regular class

Marcono1234:
Relates to #252

Fixes empty enums (e.g. `enum EnumTest { }`) compiled with Java 15+ being dumped as regular class.
I have simply adjusted `getJava15Values` to account for a `NewObjectArray` expression. Feel free to close this pull request and implement this in a saner / cleaner way, or let me know if I should change anything.

### DIFF
--- a/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/EnumClassRewriter.java
+++ b/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/EnumClassRewriter.java
@@ -87,7 +87,7 @@ public class EnumClassRewriter {
 
     private boolean rewrite() {
         // Ok, it's an enum....
-        // Verify the static initialiser - for each instance of enm created in there, make sure
+        // Verify the static initialiser - for each instance of enum created in there, make sure
         // it's a public static enum member.
         // Verify that the values field is also valid.
         // Nop out this code from the static initialiser.  There MAY be code left, if the enum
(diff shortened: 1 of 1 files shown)


src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/EnumClassRewriter.java
```diff
@@ -416,8 +434,17 @@ boolean isValid() {
         // Return of new anonymous array.
         if (!(testArr instanceof StructuredReturn)) return null;
         Expression rv = ((StructuredReturn)testArr).getValue();
-        if (!(rv instanceof NewAnonymousArray)) return null;
```
leibnitz27, line 419: I think you could just swap rv for a temp NewAnonymousArray if it matched the newObjectArray tests below here, and the rest of the change wouldn't be needed - need to go through your change though as I think it makes stuff clearer anyway.

leibnitz27:
Sorry I took so long to get around to this!  As per comment I think there's a 2-liner change - just want to figure out if my current overcomplication is overcomplicated for a reason or if because I'm overcomplicated.

leibnitz27:
Ah - guess I was just trying a bit too hard to re-use CollectedEnumData for a disjoint purpose.

src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/EnumClassRewriter.java
Marcono1234: Do you mean replace `rv`'s value with a 'fake' `NewAnonymousArray` in case it matches `new EnumType[0]`, and then revert the changes this pull request makes to `CollectedEnumData`?
Do you still want this to be changed, or is it fine like this?

leibnitz27: Nah, I was just thinking out loud.  Merged this, thanks.

PR #293, opened 2022-03-20
Patch formatting

nbauma109:
Format try/catch/finally as follows:

```java
try {

} catch (Exception e) {

} finally {

}
```

rather than:

```java
try {

}
catch (Exception e) {

}
finally {

}
```

It's not just for esthetics and conventions, it makes the output more ready for line number realignment.
In Enhanced Class Decompiler (ECD) plugin for Eclipse, the output is reworked for line number realignment and it doesn't look nice.
This will help produce a nicer output.

Example of the current messy output of ECD:

```java
/*    */          public class Test {
/*    */              public static void main(String[] args) {
/*    */                  try {
/*  5 */                      System.out.println("try");
/*  6 */                      if (args.length == 0) {
/*  7 */                          System.out.println("no args");
/*    */                      } else {
/*  9 */                          System.out.println(args.length + " args");                     }
/* 10 */                      if (args.length == 1) {
/* 11 */                          System.out.println("1 arg");                     }
/* 12 */                      if (args.length == 2) {
/* 13 */                          System.out.println("2 args");
/*    */                      }                 }                 catch (Exception e) {
/* 15 */                      System.out.println("catch ex");
/*    */                  }                 catch (Error e) {
/* 17 */                      System.out.println("catch err");
/*    */                  }                 finally {
/* 19 */                      System.out.println("finally");
/*    */                  }
/*    */              }
/*    */          }
```

I have thought of doing that directly in ECD, because I have the same issue with Procyon, but as it's also the java standard, I think the decompiler can produce that output directly.


src/org/benf/cfr/reader/bytecode/analysis/structured/statement/StructuredTry.java
```diff
@@ -77,8 +78,15 @@ public Dumper dump(Dumper dumper) {
             dumper.separator(")");
         }
         tryBlock.dump(dumper);
-        for (Op04StructuredStatement catchBlock : catchBlocks) {
+        dumper.removePendingCarriageReturn();
```
leibnitz27, line 81: This can be achieved just by having dumper.removePendingCarriageReturn().separator(" ") 

Here

src/org/benf/cfr/reader/bytecode/analysis/structured/statement/StructuredTry.java
```diff
@@ -77,8 +78,15 @@ public Dumper dump(Dumper dumper) {
...
+        for (Iterator<Op04StructuredStatement> it = catchBlocks.iterator(); it.hasNext();) {
+            Op04StructuredStatement catchBlock = it.next();
             catchBlock.dump(dumper);
+            if (it.hasNext() || finallyBlock != null) {
+                dumper.removePendingCarriageReturn();
+                dumper.print(' ');
+            }
         }
         if (finallyBlock != null) {
             finallyBlock.dump(dumper);
```
leibnitz27, line 92: and here

leibnitz27:
Was going to go off on a little rant about style guides being subjective, but you win on points because this IS actually in the *official* sun style guide (which was last published in 1999 ;) )
https://www.oracle.com/java/technologies/javase/codeconventions-contents.html

src/org/benf/cfr/reader/bytecode/analysis/structured/statement/StructuredTry.java
```diff
@@ -77,8 +78,13 @@ public Dumper dump(Dumper dumper) {
             dumper.separator(")");
         }
         tryBlock.dump(dumper);
-        for (Op04StructuredStatement catchBlock : catchBlocks) {
+        dumper.removePendingCarriageReturn().separator(" ");
```
leibnitz27, line 81: Lol - What I meant was - there's no need to mess around with iterators - your code is equivalent to

        tryBlock.dump(dumper);
        for (Op04StructuredStatement catchBlock : catchBlocks) {
            dumper.removePendingCarriageReturn().separator(" ");
            catchBlock.dump(dumper);
        }
        if (finallyBlock != null) {
            dumper.removePendingCarriageReturn().separator(" ");
            finallyBlock.dump(dumper);
        }

PR #300, opened 2022-04-03
Fix `& Serializable` casts being omitted for method reference expressions

Marcono1234:
When a lambda or method reference expression is cast with an intersection type consisting of `Serializable` (e.g. `(Runnable & Serializable)`) it makes the lambda or method reference serializable. That cast must therefore not be omitted.

Previously CFR handled this correctly for lambda expressions, but for method references it erroneously omitted the cast. This pull request tries to fix this. I am however not sure if this is the correct approach, maybe there is an underlying bug in how CFR determines whether an intersection type cast may be omitted.

#### Source code
```java
import java.io.Serializable;
import java.util.function.IntFunction;

class LambdaSerialization {
    private static void doSomething() {}
    private static void doSomethingWithParam(int i) {}
    
    public static void main(String[] args) throws Exception {
        Runnable r;
        r = (Runnable & Serializable) () -> doSomethingWithParam(0);
        r = (Runnable & Serializable) () -> doSomething();
        r = (Runnable & Serializable) LambdaSerialization::doSomething;
        r = (Runnable & Serializable) LambdaSerialization::new;
        
        Object o;
        o = (IntFunction<String[]> & Serializable) String[]::new;
    }
}
```

#### Output
Before (CFR 0.153-SNAPSHOT (c69ed17)):
```java
Runnable runnable = (Runnable & Serializable)() -> LambdaSerialization.doSomethingWithParam(0);
runnable = (Runnable & Serializable)() -> LambdaSerialization.doSomething();
runnable = LambdaSerialization::doSomething;
runnable = LambdaSerialization::new;
IntFunction intFunction = String[]::new;
```

After:
```java
Runnable runnable = (Runnable & Serializable)() -> LambdaSerialization.doSomethingWithParam(0);
runnable = (Runnable & Serializable)() -> LambdaSerialization.doSomething();
runnable = (Runnable & Serializable)LambdaSerialization::doSomething;
runnable = (Runnable & Serializable)LambdaSerialization::new;
IntFunction intFunction = (IntFunction<String[]> & Serializable)(String[]::new);
```

### DIFF
--- a/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/LambdaRewriter.java
+++ b/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/LambdaRewriter.java
@@ -33,6 +33,7 @@ import org.benf.cfr.reader.bytecode.analysis.structured.statement.StructuredRetu
 import org.benf.cfr.reader.bytecode.analysis.types.DynamicInvokeType;
 import org.benf.cfr.reader.bytecode.analysis.types.GenericTypeBinder;
 import org.benf.cfr.reader.bytecode.analysis.types.JavaGenericRefTypeInstance;
+import org.benf.cfr.reader.bytecode.analysis.types.JavaIntersectionTypeInstance;
 import org.benf.cfr.reader.bytecode.analysis.types.JavaRefTypeInstance;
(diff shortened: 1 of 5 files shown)


leibnitz27:
Introduces a minor regression in LambdaIntersectionTraitTest, will fix on master.

PR #320, opened 2022-06-01
increasing visibility to build an option panel

nbauma109:
I've already just built an option panel in a GUI and this  would be useful to have these methods as public for providing tooltip texts on the options instead of calling setAccessible(true) by reflection, so I thought it might make more sense to send this PR.

### DIFF
--- a/src/org/benf/cfr/reader/util/getopt/PermittedOptionProvider.java
+++ b/src/org/benf/cfr/reader/util/getopt/PermittedOptionProvider.java
@@ -28,7 +28,7 @@ public interface PermittedOptionProvider {
             return name;
         }
 
-        OptionDecoderParam<X, InputType> getFn() {
+        public OptionDecoderParam<X, InputType> getFn() {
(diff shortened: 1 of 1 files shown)


nbauma109:
Unsatisfying code on my side is
https://github.com/nbauma109/jd-gui-duo/blob/bd368eacc340db30c89f073832d615cffd535ccd/services/src/main/java/org/jd/gui/service/preferencespanel/CFRDecompilerPreferencesProvider.java#L49
https://github.com/nbauma109/jd-gui-duo/blob/bd368eacc340db30c89f073832d615cffd535ccd/services/src/main/java/org/jd/gui/service/preferencespanel/CFRDecompilerPreferencesProvider.java#L54

nbauma109:
You may want to have a look at this GUI, it will help you to visualize decompiler bugs quickly

Example
image

Col-E:
As I also use reflection to gain access to option values, this would be nice to have.

PR #326, opened 2022-07-03
Dump varargs parameter annotations

Marcono1234:
Previously no varargs parameter annotations were dumped. This can be seen with the following test source (feel free to add it to the tests):
```java
import static java.lang.annotation.RetentionPolicy.RUNTIME;

import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.Target;

class VarargsAnnotationTest {
    @Retention(RUNTIME)
    @Target(ElementType.PARAMETER)
    @interface Parameter { }
    
    @Retention(RUNTIME)
    @Target(ElementType.TYPE_USE)
    @interface TypeUse {
        int value();
    }
    
    void varargs1(@Parameter @TypeUse(1) String ... a) {}
    void varargs2(@Parameter String @TypeUse(2) ... a) {}
    void varargs3(@Parameter @TypeUse(1) String @TypeUse(2) ... a) {}
    
    void varargsArray(@Parameter @TypeUse(1) String @TypeUse(2) [] @TypeUse(3) ... a) {}
    
    void array(@Parameter @TypeUse(1) String @TypeUse(2) [] a) {}
    
    void varargsParam(@Parameter String... a) {}
    
    void arrayParam(@Parameter String[] a) {}
}
```

### DIFF
--- a/src/org/benf/cfr/reader/bytecode/analysis/types/MethodPrototypeAnnotationsHelper.java
+++ b/src/org/benf/cfr/reader/bytecode/analysis/types/MethodPrototypeAnnotationsHelper.java
@@ -80,7 +80,15 @@ public class MethodPrototypeAnnotationsHelper {
         return typeEntries;
     }
 
-    void dumpParamType(JavaTypeInstance arg, final int paramIdx, Dumper d) {
+    void dumpParamType(JavaTypeInstance arg, int paramIdx, Dumper d) {
+        dumpParamType(arg, paramIdx, false, d);
+    }
+
+    void dumpVarargsParamType(JavaArrayTypeInstance arg, int paramIdx, Dumper d) {
+        dumpParamType(arg, paramIdx, true, d);
+    }
+
+    private void dumpParamType(JavaTypeInstance arg, final int paramIdx, boolean isVarargs, Dumper d) {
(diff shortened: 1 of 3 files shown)


src/org/benf/cfr/reader/bytecode/analysis/types/MethodPrototypeAnnotationsHelper.java
```diff
@@ -96,13 +104,22 @@ void dumpParamType(JavaTypeInstance arg, final int paramIdx, Dumper d) {
...
+            }
         } else {
             JavaAnnotatedTypeInstance jat = arg.getAnnotatedInstance();
             DecompilerComments comments = new DecompilerComments();
             TypeAnnotationHelper.apply(jat, typeAnnotationsToDump, comments);
             d.dump(comments);
-            d.dump(jat);
+
+            if (isVarargs) {
+                ((JavaArrayTypeInstance.Annotated) jat).dumpVarargs(d);
```
leibnitz27, line 119: I know this is the only type it could be, but twitchy about having this cast here

leibnitz27:
As per the comment elsewhere - I think it would be nicer if we didn't rely on explicit casts - have tweaked as per d6f6758ee900ae1a1fffebe2d75c69ce21237b2a

PR #327, opened 2022-07-10
Fix compact canonical record constructor not using correct parameter names

### LINKED ISSUE #319: Canonical record constructor performing assignment uses wrong variable name
# CFR version
c69ed17cebd473773b0f2d34cd727e1653b30537

# Compiler
javac 17

# Description
When the canonical constructor of a record performs an operation on one of the implicit parameters and assigns the value to the record component, CFR might erroneously use a CFR-generated parameter name instead of the component name.

# Example
Source code:
```java
record Record2(int test, String myString) {
    Record2 {
        myString = myString.trim();
    }
}
```

Decompiled output:
```java
record Record2(int test, String myString) {
    Record2 {
        string = string.trim();
    }
}
```
:x: The output uses the implicit parameter with a CFR-generated name (`string`) which would cause a compilation failure.

Marcono1234:
Fixes #319

### DIFF
--- a/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/RecordRewriter.java
+++ b/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/RecordRewriter.java
@@ -312,6 +312,10 @@ public class RecordRewriter {
         }
     }
 
+    /**
+     * Tries to remove implicit field assignments from this canonical record constructor, turning
+     * it into a compact canonical constructor. Returns {@code true} if successful.
+     */
     private static boolean removeImplicitAssignments(Method canonicalCons, List<ClassFileField> instances, JavaRefTypeInstance thisType) {
         if (canonicalCons.getCodeAttribute() == null) return false;
         Op04StructuredStatement code = canonicalCons.getAnalysis();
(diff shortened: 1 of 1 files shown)


src/org/benf/cfr/reader/bytecode/analysis/opgraph/op4rewriters/RecordRewriter.java
```diff
@@ -333,18 +337,22 @@ private static boolean removeImplicitAssignments(Method canonicalCons, List<Clas
...
             instances.set(idx, null);
 
-            Expression rhs = ((StructuredAssignment) statement).getRvalue();
+            Expression rhs = assignment.getRvalue();
             if (!(rhs instanceof LValueExpression)) break;
             LValue rlv = ((LValueExpression) rhs).getLValue();
             LocalVariable expected = args.get(idx);
             if (rlv != expected) break;
+            // Implicit parameter must use the same name as the field
+            expected.getName().forceName(field.getFieldName());
```
Marcono1234, line 355: Might have to be moved down outside the loop to be sure that this is actually a compact canonical constructor. But probably does not hurt to have it here either, since the condition in the line before already made sure that parameter and field match.

Marcono1234:
@leibnitz27, if possible could you please have a look at this? The issue still exists in the latest SNAPSHOT version.

PR #377, opened 2026-04-02
Add Uploading JARS to artifacts

a-crazy-pencil:
It adds the ability to upload all the JARs to artifacts.


a-crazy-pencil (commented):
Ready

.github/workflows/maven.yml
```diff
@@ -30,7 +30,35 @@ jobs:
...
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v7
       with:
         name: decompilation-test-failures-diff
         path: target/cfr-test-failures-diff
+
+    - name: Upload main jar to artifact
+      uses: actions/upload-artifact@v7
+      with:
+        name: CFR Main JAR
```
Marcono1234: Are you sure this combination of `name: ...` + `archive: false` works as desired? The [documentation](https://github.com/actions/upload-artifact/releases/tag/v7.0.0) says in that case `name` is ignored.
So I think `archive: false` should suffice.

(same for the others below as well)

pizzapi2012: it does work as i desire.

pizzapi2012: @Marcono1234

pizzapi2012: Wait i know what you mean i will remove the extra name tags

a-crazy-pencil (commented):
Ready

Marcono1234:
@leibnitz27, what do you think? I think this could be quite useful if users want to try out the latest SNAPSHOT version without building CFR themselves (which requires them to clone the repository and set up a JDK which still supports Java 6 as target).
(Though I am not sure how useful the `-sources.jar` and `-javadoc.jar` file will be.)

The artifacts will only be [retained for 90 days](https://github.com/actions/upload-artifact/tree/v7.0.0#retention-period), but that might be good enough?

a-crazy-pencil:
i can add releases to that they last longer

.github/workflows/maven.yml
```diff
@@ -30,7 +30,32 @@ jobs:
...
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v7
       with:
         name: decompilation-test-failures-diff
         path: target/cfr-test-failures-diff
+
+    - name: Upload main jar to artifact
+      uses: actions/upload-artifact@v7
+      with:
+        path: target/*[^-sources][^-javadoc].jar
```
leibnitz27: Glob documentation is sparse, but this is 2 successive posix character classes - I don't think this is doing what you intend.  ( https://github.com/isaacs/minimatch/blob/main/src/brace-expressions.ts#L88-L91 ) 

(i.e. it discards any jar where the last 2 characters are in {-,s,o,u,r,c,e,s}{-,j,a,v,a,d,o,c}.   

Which does coindidentally work, but I don't think that's the intent.

a-crazy-pencil: how would I fix it then?

a-crazy-pencil: @leibnitz27

leibnitz27: - name: Upload main jar to artifact
        uses: actions/upload-artifact@v7
        with:
          path: target/cfr-*-SNAPSHOT.jar
          archive: false

      - name: Upload sources jar to artifact
        uses: actions/upload-artifact@v7
        with:
          path: target/cfr-*-SNAPSHOT-sources.jar
          archive: false

      - name: Upload javadoc jar to artifact
        uses: actions/upload-artifact@v7
        with:
          path: target/cfr-*-SNAPSHOT-javadoc.jar
          archive: false

would seem simpler?

a-crazy-pencil: Made the changes ready to merge

leibnitz27:
Merged, thanks.

PR #378, opened 2026-04-05
Fix wrong module attribute length

### LINKED ISSUE #335: module-info cannot be decompiled after packing into jar
# CFR version
0.152

# Compiler
javac 17.0.3 from Azul Zulu (I need JavaFX)

# Description
I patched in module info so I can use jlink according to [this stackoverflow comment](https://stackoverflow.com/questions/47222226#comment82872402_47222302)
The resulting module-info.class can be decompiled before packing, but not after packing and extracting O_o
```
Can't load the class specified:
org.benf.cfr.reader.util.CannotLoadClassException: .\module-info.class - org.benf.cfr.reader.util.ConfusedCFRException: Attempt to fetch element 0 from constant pool
```

# Example
[module-info java + classes.zip](https://github.com/leibnitz27/cfr/files/11101701/module-info.java.%2B.classes.zip)

Marcono1234:
Fixes #335

The issue was probably not noticeable normally because often there is no subsequent attribute after the `Module` one. However, for the `module-info.class` provided in that issue, there was a subsequent attribute, and the incorrect length caused the read position to be incorrect.

Side note: In general it might be good to refactor and simplify the attribute reading logic? It seems the [structure is always the same](https://docs.oracle.com/javase/specs/jvms/se25/html/jvms-4.html#jvms-4.7); the length is not something specific to a certain attribute. So it might be better to move that outside the attribute reading and remove the overridden `getRawByteLength` method from all attribute classes.
It seems all other attribute implementations return the correct `6 + length`, but this code repetition is rather error-prone.

### DIFF
--- a/src/org/benf/cfr/reader/entities/attributes/AttributeModule.java
+++ b/src/org/benf/cfr/reader/entities/attributes/AttributeModule.java
@@ -264,7 +264,7 @@ public class AttributeModule extends Attribute {
 
     @Override
     public long getRawByteLength() {
-        return OFFSET_OF_MODULE_FLAGS + length;
+        return OFFSET_OF_MODULE_NAME + length;
     }
 
     @Override


leibnitz27:
Thanks!  Y - you're right, that code is *way* too repetitive.... :(

PR #379, opened 2026-04-05
Fix generic type clash not being detected

### LINKED ISSUE #350: Incorrect Comparable Expression Produced by Ternary Expression
# CFR version
CFR 0.153-SNAPSHOT (24c7433-dirty)

# Compiler
javac 11.0.20.1

# Description
The ternary expression may be decompiled into an incorrect Comparable expression.

# Example
The example code is as follows:

```java
class T1 {
    public Object foo() {
        Boolean b1 = false;
        Boolean b2 = true;
        final Object o = ((b1) ? -33.998 : b2);
        return o;
    }
}
```

The decompiled code is as follows:

```java
/*
 * Decompiled with CFR 0.153-SNAPSHOT (24c7433-dirty).
 */
class T1 {
    T1() {
    }

    public Object foo() {
        Boolean bl = false;
        Boolean bl2 = true;
        Comparable<Boolean> comparable = bl != false ? (Comparable<Double>)Double.valueOf(-33.998) : (Comparable<Double>)bl2;
        return comparable;
    }
}
```

Compiling the code results in the following compilation error:

```txt
./T1.java:11: error: incompatible types: bad type in conditional expression
        Comparable<Boolean> comparable = bl != false ? (Comparable<Double>)Double.valueOf(-33.998) : (Comparable<Double>)bl2;
                                                       ^
    Comparable<Double> cannot be converted to Comparable<Boolean>
./T1.java:11: error: incompatible types: Boolean cannot be converted to Comparable<Double>
        Comparable<Boolean> comparable = bl != false ? (Comparable<Double>)Double.valueOf(-33.998) : (Comparable<Double>)bl2;
                                                                                                                         ^
2 errors
```

Marcono1234:
Fixes #350

The problem was that the code previously only considered the raw supertype but did not check for parameterized type incompatibilities.

Note: This might lead to worse decompilation results which need more casts, but the results should be more correct (?).

### DIFF
--- a/src/org/benf/cfr/reader/bytecode/analysis/types/discovery/InferredJavaType.java
+++ b/src/org/benf/cfr/reader/bytecode/analysis/types/discovery/InferredJavaType.java
@@ -149,7 +149,22 @@ public class InferredJavaType {
                     continue;
                 }
                 Map<? extends JavaTypeInstance, JavaGenericRefTypeInstance> boundSupers = otherSupers.getBoundSuperClasses();
-                matches.keySet().retainAll(boundSupers.keySet());
+
+                Iterator<Map.Entry<JavaTypeInstance, JavaGenericRefTypeInstance>> matchesIter = matches.entrySet().iterator();
+                while (matchesIter.hasNext()) {
+                    Map.Entry<JavaTypeInstance, JavaGenericRefTypeInstance> entry = matchesIter.next();
+                    if (!boundSupers.containsKey(entry.getKey())) {
+                        matchesIter.remove();
+                    } else {
+                        JavaGenericRefTypeInstance matchGeneric = entry.getValue();
+                        JavaGenericRefTypeInstance otherGeneric = boundSupers.get(entry.getKey());
+
+                        if (!MiscUtils.objectsEquals(matchGeneric, otherGeneric)) {
+                            // TODO: Try to determine common generic supertype?
+                            matchesIter.remove();
+                        }
+                    }
+                }
             }
             return matches;
         }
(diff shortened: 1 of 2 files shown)


src/org/benf/cfr/reader/util/MiscUtils.java
```diff
@@ -48,6 +49,10 @@ public boolean test(String in) {
     public static void handyBreakPoint() {
     }
 
+    public static boolean objectsEquals(@Nullable Object o1, @Nullable Object o2) {
+        return o1 == o2 || o1 != null && o1.equals(o2);
+    }
```
Marcono1234, line 54: This is effectively `java.util.Objects#equals`, which is unfortunately not available.

I hope this was not defined somewhere else already and I overlooked it.

src/org/benf/cfr/reader/bytecode/analysis/types/discovery/InferredJavaType.java
```diff
@@ -149,7 +149,22 @@ private static Map<JavaTypeInstance, JavaGenericRefTypeInstance> getMatches(List
...
+                    Map.Entry<JavaTypeInstance, JavaGenericRefTypeInstance> entry = matchesIter.next();
+                    if (!boundSupers.containsKey(entry.getKey())) {
+                        matchesIter.remove();
+                    } else {
+                        JavaGenericRefTypeInstance matchGeneric = entry.getValue();
+                        JavaGenericRefTypeInstance otherGeneric = boundSupers.get(entry.getKey());
+
+                        if (!MiscUtils.objectsEquals(matchGeneric, otherGeneric)) {
+                            // TODO: Try to determine common generic supertype?
+                            matchesIter.remove();
```
Marcono1234, line 164: I created a prototype locally for determining the common parameterized type, but I am not sure if the additional complexity is worth it, and it causes another issue.

While this does lead to a potentially more specific type, CFR seems to create incompatible casts for the ternary branches then.
Decompiled example:
```java
Comparable<Constable> comparable = bl != false ? (Constable)Double.valueOf(-33.998) : (Constable)bl2
```
It uses `Comparable<Constable>` which is potentially more specific[^1], but the `(Constable)` casts are incorrect because `Constable` does not extend `Comparable`.

[^1]: It is not actually more specific because `Comparable` and `Constable` are unrelated types.

leibnitz27:
Thanks.   Sorry for delay, there's some other gnarly edge cases with flowing clashes that sprung out of testing this!

PR #383, opened 2026-05-12
Remove redundant null check for `instanceof` patterns

Marcono1234:
`CondenseConditionals#condenseInstanceOfAssign` introduces a redundant null-check (apparently to lift the assignment into the `if` condition?) which ends up in the decompiled output:
https://github.com/leibnitz27/cfr/blob/a0d8fbfd4ef24ddbc4f54db500267a44f0b56307/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op3rewriters/CondenseConditionals.java#L506-L509

This PR here tries to detect that and remove it again.

Example (JDK 21):
```java
class InstanceOfTest {
    int m(Object o1, Object o2) {
        if (o1 instanceof String s1 && o2 instanceof String s2) {
            return s1.compareTo(s2);
        }
        return 0;
    }
}
```

Decompiled:
```diff
- if (object instanceof String var3_3 && null != var3_3 && object2 instanceof String) {
+ if (object instanceof String var3_3 && object2 instanceof String) {
```

### DIFF
--- a/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op3rewriters/CondenseConditionals.java
+++ b/src/org/benf/cfr/reader/bytecode/analysis/opgraph/op3rewriters/CondenseConditionals.java
@@ -538,6 +538,20 @@ public class CondenseConditionals {
         return effect;
     }
 
+    /**
+     * Returns whether {@code cond} is a redundant null-check for the known non-null {@code nonNullVar}.
+     *
+     * <p>This only covers the artificially introduced null-check added by {@link #condenseInstanceOfAssign}.
+     */
+    public static boolean isRedundantInstanceOfNullCheck(ConditionalExpression cond, LValue nonNullVar) {
+        if (cond instanceof ComparisonOperation) {
+            ComparisonOperation comp = (ComparisonOperation) cond;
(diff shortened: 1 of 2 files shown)


Marcono1234:
Have marked this as Draft again. This is really not an ideal solution; ideally `CondenseConditionals#condenseInstanceOfAssign` would not introduce the null-check in the first place, but I don't know if that is possible.

What do you think? Though after all, the null-check is not 'incorrect' (and is probably side-effect free); it is just redundant. So I could understand if you decide to keep it as it is and close this PR unmerged.


Even with the changes of this PR this will only remove the null-check if an `instanceof` pattern is emitted, otherwise the null-check remains. For example if you negate the example from above:
```java
class InstanceOfTest {
    int m(Object o1, Object o2) {
        if (!(o1 instanceof String s1 && o2 instanceof String s2)) {
            return 1;
        }
        return 0;
    }
}
```

Decompiled:
```java
if (!(object instanceof String) || null == (string = (String)object) || !(object2 instanceof String)) {
```

leibnitz27:
Hey - I have a fairly large pipeline change which is continuing to improve most of this, will look again after merged.

leibnitz27:
we're good now (Added this as an explicit test)

PR #384, opened 2026-05-12
Simplify reading of attributes

Marcono1234:
Follow-up for #378

Simplifies the code by moving the common attribute reading logic (name + length) to `AttributeFactory` instead of duplicating it for every attribute class.

### DIFF
--- a/src/org/benf/cfr/reader/entities/attributes/AttributeLineNumberTable.java
+++ b/src/org/benf/cfr/reader/entities/attributes/AttributeLineNumberTable.java
@@ -9,25 +9,19 @@ import java.util.TreeMap;
 public class AttributeLineNumberTable extends Attribute {
     public static final String ATTRIBUTE_NAME = "LineNumberTable";
 
-    private static final long OFFSET_OF_ATTRIBUTE_LENGTH = 2;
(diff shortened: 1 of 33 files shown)


src/org/benf/cfr/reader/entities/attributes/AttributeLineNumberTable.java
```diff
@@ -9,25 +9,19 @@
...
+    private static final long OFFSET_OF_ENTRIES = 2;
 
-    private final int length;
     private final NavigableMap<Integer, Integer> entries = new TreeMap<Integer, Integer>();
 
 
     public AttributeLineNumberTable(ByteData raw) {
-        this.length = raw.getS4At(OFFSET_OF_ATTRIBUTE_LENGTH);
         int numLineNumbers = raw.getU2At(OFFSET_OF_ENTRY_COUNT);
-        if (numLineNumbers * 2 <= length) {
```
Marcono1234, line 24: Have removed this length check. It seems this was the only attribute class doing this, and I assume the JVM might not permit such a malformed attribute entry in the first place (?).

Please let me know if I should add it back.

leibnitz27:
Thanks - that stuff was rather..... evolved!

